Issue #160, opened 2014-12-17
Semi-boilerplate-y solution for lens creation


ncreep:
Hi,

I thought that it may be possible to create lenses with a bit less boilerplate than what `Lenser` currently provides, but without resorting to macro annotations.

I've implemented what seems to be a working solution, and it can be found in [this](https://github.com/ncreep/Monocle/commit/0733af386af2dae142911bd3627dbdf7e87de962) commit.

The upshot is that it is now possible to write:

``` scala
case class Point(x: Int, y: Int)

val lenses = CaseClassLenser.getLenses[Point]
import lenses._

val update = x.modify(_ + 100) compose y.set(7)
update(Point(1, 2)) ==== Point(101, 7)
```

This approach has the following advantages:
- Less boilerplate than `Lenser` 
- No need for macro annotations or the macro paradise plugin
- Works even when the target case class cannot be edited

But also the following cons:
- It's a whitebox macro that generates a structural type, so support from IDEs might be a bit flaky (though auto-completion seems to work)
- To avoid reflective calls, the macro inlines the results at the call sites, creating some compilation overhead (more details in the documentation)

If this may be useful for the project, I can submit a pull request.

PR #164, opened 2014-12-22
Semi-boilerplate-y solution for lens creation

ncreep:
As discussed in #160


julien-truffaut:
as discussed in #160, this PR is postponed until we get information about the future of whitebox macro. 

Thanks @ncreep

PR #170, opened 2015-01-19
Update Iso definition and add Iso macro

julien-truffaut:


NightRa:
Summing up:
1. Isoer seems like a bad name. (consistant naming should not go too far)
2. The macro should check if it's operating on a case class
3. The macro should verify there is only one field.

Other than that, everything looks good!

julien-truffaut:
@NightRa I agree it is not a good name, I welcome suggestions!

I added checks to produce useful error message. Next step: some doc

Issue #171, opened 2015-01-21
Makes all Optics abstract class and remove any public functions in constructor


julien-truffaut:
I propose to make the following transformation to all Optics:

```scala
abstract class PPrism[S, T, A, B] private[monocle](val getOrModify: S => T \/ A, val reverseGet: B => T){...}

// to

abstract class PPrism[S, T, A, B] private[monocle]{
  def getOrModify(s: S): T \/ A
  def reverseGet(b: B): T
}
```

benefits:
* we can make some optimizations e.g. efficient macro impl
* non committing implementation: we should be able to change the implementation without a binary change. At the moment we expose that `getOrModify` is a `Function1` instead of a method which is an implementation detail

PR #173, opened 2015-01-21
#171 make all optics (except Traversal and Fold) an abstract class

julien-truffaut:


core/src/main/scala/monocle/Lens.scala
```diff
@@ -78,8 +78,8 @@ abstract class PLens[S, T, A, B] private[monocle]{ self =>
       def get(s: S): C =
         other.get(self.get(s))
 
-      def set(b: D): S => T =
-        self.modify(other.set(b))
+      def set(d: D): S => T =
+        self.modify(other.set(d))
```
NightRa: Thought of mentioning this when I prepared for my talk :)
The PR looks alright overall.

PR #174, opened 2015-01-25
add Compose, Category, Split and Choice instance for all optics

julien-truffaut:


core/src/main/scala/monocle/Lens.scala
```diff
@@ -198,4 +198,12 @@ object Lens {
...
   def apply[S, A](get: S => A)(set: A => S => S): Lens[S, A] =
     PLens(get)(set)
+
+  implicit val lensCategory: Category[Lens] = new Category[Lens] {
+    def id[A]: Lens[A, A] =
+      Iso.id[A].asLens
+
+    def compose[A, B, C](f: Lens[B, C], g: Lens[A, B]): Lens[A, C] =
+      g composeLens f
+  }
```
NightRa: How can we supply a Category instance for PLens?
It's just function composition, but with restricted types.

julien-truffaut: we can't, Compose.compose take 3 type parameters but PLens.compose take 6. We need to restrict `PLens` to `Lens` in order to make an instance of `Category`

julien-truffaut:
TODO: add `Compose`, `Split` and `Choice` instances

Issue #172, opened 2015-01-21
Create GenPrism macro


julien-truffaut:
```scala
sealed trait Foo
 
case class A(i: Int) extends Foo
case class B(x: Double, y: Double) extends Foo
case object C extends Foo
 
// 1st option:
 
GenPrism[Foo, A]: Prism[Foo, A]
GenPrism[Foo, B]: Prism[Foo, B]
GenPrism[Foo, C]: Prism[Foo, C]
 
// 2nd option:
 
GenPrism[Foo, A, Int]: Prism[Foo, Int]
GenPrism[Foo, B, (Double, Double)]: Prism[Foo, (Double, Double)]
GenPrism[Foo, C, Unit]: Prism[Foo, Unit]
```

If one day we have a macro to generate an `Iso` for any case class, we could write option 2 in term of 1:
```scala
GenPrism[Foo, B] composeIso GenIso[B, (Double, Double)]
```

At the moment, I favor option 2 because scala ADT implementation via sub typing causes some pain when working with type classes.

PR #176, opened 2015-01-31
Add GenPrism macro

julien-truffaut:


julien-truffaut:
solving #172

NightRa:
Missing check for an inheritance relation between the classes.
Otherwise, looks good.

macro/src/main/scala/monocle/macros/GenPrism.scala
```diff
@@ -0,0 +1,38 @@
+package monocle.macros
+
+import monocle.Prism
+import monocle.macros.internal.MacrosCompatibility
+
+
+
+
+object GenPrism {
+  def apply[S, A <: S]: Prism[S, A] = macro GenPrismImpl.genPrism_impl[S, A]
```
julien-truffaut: @NightRa I am doing the check here

NightRa: Oh... alright :)
You should probably put a comment saying that.

julien-truffaut: good point

Issue #153, opened 2014-12-04
add optics for scalaz.Validation


julien-truffaut:

mingp:
Hello @julien-truffaut. I saw your comment on Reddit /r/scala about getting started with contributing to Monocle with issues labeled "padawan friendly". Could you please clarify a few things about this issue and about Monocle in general?

* What is the recommended way to familiarize one's self with the Monocle code base? Are there any particular classes/methods/files to look at for examples?
* Approximately what components should be implemented for this issue? What documentation/tests should be updated to correspond to such?
* Any particular examples of existing code similar to what should be written for this issue which could be used as example/inspiration?

Thank you very much in advance!

julien-truffaut:
Hi @mingp 

Sure, it might be a bit intimidating to dive into the code base. What's important is that you familiarize yourselves with a few Optics and how they relate to each other, for example I would have a look a `Lens`, `Prism` and `Optional` (module core, package monocle). Basically each of this optics is a pair of function (get / set, getMaybe / reverseGet and getMaybe / set respectively). The key point is that optics compose with each other to create new optics, e.g. if you compose a `Lens` with a `Optional` you get an `Optional` (cf diagram in readme).

Once you have done that you can have a look at `Maybe` and `Either` optics instances: https://github.com/julien-truffaut/Monocle/blob/master/core/src/main/scala/monocle/std/Maybe.scala and https://github.com/julien-truffaut/Monocle/blob/master/core/src/main/scala/monocle/std/Either.scala

This issue is about implementing similar optics but for `scalaz.Validation` (kind of an `Either` that accumulates failure). You can start with the one I described i.e. success, failure and validationToMaybe.

For testing, we generally use optics laws to verify our implementation. In other word, you have almost nothing to do, here are the tests for `Either`: https://github.com/julien-truffaut/Monocle/blob/master/test/src/test/scala/monocle/std/EitherSpec.scala

Don't hesitate if you have any questions, you can reach me or other people familiar with the library on irc or gitter (once I figure out how to access it fro my job ^^)

anakos:
Hello-

Seeing as how the previous PR has been closed and remains unmerged, I'm going to take a stab at this if no one else is working on it.

Alex

PR #178, opened 2015-02-02
Add optics for `scalaz.Validation` #153

mingp:
This is my initial attempt at implementing #153, "add optics for `scalaz.Validation".

This PR includes:
* `def success[E, A]: Prism[Validation[E, A], A]`.
* `def failure[E, A]: Prism[Validation[E, A], E]`,
* `implicit def validationArbitrary[E: Arbitrary, A: Arbitrary]: Arbitrary[Validation[E, A]]`, as was needed for `scalacheck` to work correctly with `scalaz.Validation`.

Some questions I still have:
* Are the names `success` and `failure` potentially too collision-prone? Should they be renamed or gain an additional prefix, e.g. `scalazSuccess` and `scalazFailure`?
* What is the meaning of `def validationToMaybe[E, A]: Optional[Validation[E, A], Maybe[A]]`? What should it produce in the `Success` case? What should it produce in the `Failure` case?
* In tests, I needed to explicitly refer to `validation.success` and `validation.failure`. I notice that the linked example `EitherSpec` directly refers to `stdLeft` and `stdRight`. Do I need to add another alias somewhere for `success` and `failure`?

Thanks in advance for clarifications!

### DIFF
--- a/core/src/main/scala/monocle/std/Validation.scala
+++ b/core/src/main/scala/monocle/std/Validation.scala
@@ -1,0 +1,20 @@
+package monocle.std
+
+import monocle.{Prism}
+import scalaz.{Validation, Maybe, Success, Failure}
+import Maybe.{Just, Empty}
+
+object validation extends ValidationFunctions
+
+trait ValidationFunctions {
+
+  def success[E, A]: Prism[Validation[E, A], A] =
+    Prism[Validation[E, A], A](_.toMaybe)(Success.apply)
+
+  def failure[E, A]: Prism[Validation[E, A], E] =
+    Prism[Validation[E, A], E]{
+      case Success(a) => Empty()
+      case Failure(e) => Just(e)
+    }(Failure.apply)
+
+}
--- a/core/src/main/scala/monocle/std/StdInstances.scala
+++ b/core/src/main/scala/monocle/std/StdInstances.scala
@@ -31,3 +31,4 @@ trait StdInstances
   with    NonEmptyListInstances
   with    OneAndInstances
   with    TreeFunctions with TreeInstances
+  with    ValidationFunctions
--- a/test/src/test/scala/monocle/TestUtil.scala
+++ b/test/src/test/scala/monocle/TestUtil.scala
@@ -88,6 +88,11 @@ object TestUtil {
     3 -> Arbitrary.arbitrary[A].map(Maybe.just(_))
   ))
 
+  implicit def validationArbitrary[E: Arbitrary, A: Arbitrary]: Arbitrary[Validation[E, A]] = Arbitrary(Gen.frequency(
+    1 -> Arbitrary.arbitrary[E].map(Failure.apply),
+    3 -> Arbitrary.arbitrary[A].map(Success.apply)
+  ))
+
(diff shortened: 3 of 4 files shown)


mingp:
Looks like the Travis CI build failed during dependency resolution/retrieval.

core/src/main/scala/monocle/std/Validation.scala
```diff
@@ -0,0 +1,20 @@
...
+object validation extends ValidationFunctions
+
+trait ValidationFunctions {
+
+  def success[E, A]: Prism[Validation[E, A], A] =
+    Prism[Validation[E, A], A](_.toMaybe)(Success.apply)
+
+  def failure[E, A]: Prism[Validation[E, A], E] =
+    Prism[Validation[E, A], E]{
+      case Success(a) => Empty()
```
julien-truffaut: if you want, you can use fold instead of pattern matching :https://github.com/scalaz/scalaz/blob/series/7.2.x/core/src/main/scala/scalaz/Validation.scala#L58-L61

julien-truffaut:
good question about name collision, for the moment I found it acceptable to `import monocle.std.validation` and then use `validation.success` but I am definitely open for better ideas! I prefixed `stdLeft` by std because they are two `Either` one from scala and one from scalaz. I wanted to support both but give the shorter name to `scalaz.\/` as it is a much better data structure.

julien-truffaut:
Here is the implementation of `Optional`: https://github.com/julien-truffaut/Monocle/blob/master/core/src/main/scala/monocle/Optional.scala

and here are its laws: https://github.com/julien-truffaut/Monocle/blob/master/law/src/main/scala/monocle/law/OptionalLaws.scala

Basically, it is a weaker `Prism` and a weaker `Lens`.

tbh, I haven't spent too much time thinking about the relation between `Validtaion` and `Maybe`, It might not be an `Optional. Feel free to drop this optic.

If you want, you can also implement an `Iso[Validation[E, A], E \/ A]` or even better a `PIso[Validation[E1, A1], Validation[E2, A2], E1 \/ A1, E2 \/ A2]`. `Validation` and `\/` are isomorphic, the only difference is their `Applicative` instance, `Validation` accumulates error while `\/` fail early

core/src/main/scala/monocle/std/Validation.scala
```diff
@@ -0,0 +1,20 @@
...
+
+import monocle.{Prism}
+import scalaz.{Validation, Maybe, Success, Failure}
+import Maybe.{Just, Empty}
+
+object validation extends ValidationFunctions
+
+trait ValidationFunctions {
+
+  def success[E, A]: Prism[Validation[E, A], A] =
```
julien-truffaut: you can also make this optic polymorphic (i.e. a `PPrism`):
```scala
def success[E, A, B]: PPrism[Validation[E, A], Validation[E, B], A, B]
```

julien-truffaut:
@mingp do you want me to merge as it is and iterate from it later on?

mingp:
@julien-truffaut Sorry for the delay. Other stuff has kept me busy lately. I would be fine either way, whether you merge this PR as-is or leave it open for adjustments.

julien-truffaut:
I am good to let it open, there is no rush

Issue #181, opened 2015-02-12
create aliases for methods returning polymorphic Optics


julien-truffaut:
```scala
// currently
def some: PPrism[Option[A], Option[B], A, B] 

// proposal
def pSome: PPrism[Option[A], Option[B], A, B] 
def some: Prism[Option[A], A] = pSome[A, A]
```

this change is motivated to get better type inference, for example:
```scala
(Option(3) &<-? some).modify(_ + 1)
```
generates the following error: 
```
type mismatch;
 found   : monocle.PPrism[Option[Int],Option[Nothing],Int,Nothing]
 required: monocle.PPrism[Option[Int],Option[Nothing],Int,B]
Note: Nothing <: B, but class PPrism is invariant in type B.
You may wish to define B as +B instead. (SLS 4.5)
  (Option(3) &<-? some).modify(_ + 1)
                  ^
```
but this compiles:
```
(Option(3) &<-? some[Int, Int]).modify(_ + 1)
```

julien-truffaut:
Assigning Meta label as there is no consensus on this issue at the moment.

PR #182, opened 2015-02-13
#181 add monomorphic aliases for Option, Maybe, Either, Disjunction, List, IList, OneAnd and Nel

julien-truffaut:


NightRa:
What is the reasoning for the split to polymorphic and monomorphic?

julien-truffaut:
In some (rather a lot of) cases type inference fails and we have to manually provide the type of an optic (e.g. `some[Int, Int]` or `disjunctionToEither[String, Int, String, Int]`). We don't have choice when we want to use polymorphic optics but in most cases you only need monomorphic ones. So my idea was to provide aliases to create monorphic optics whenever we could.

I don't know if it is worth it, I welcome all feedbacks :)

julien-truffaut:
closing waiting for more feedbacks before making a decision

PR #188, opened 2015-02-18
add macro support for monomorphic lenses for polymorphic case classes

### LINKED ISSUE #184: @Lenses for case class with type parameters
`@Lenses case class Foo[A](a: A)`
 leads to
`[error] class Foo takes type parameters`
and sometimes (I haven't figured out the necessary conditions):
`[error] Invalid annotation target: must be a case class`
(it *is* a case class!)

refried:
fix #184.  

```scala
// input:
@Lenses case class Person(name: String, age: Int)

// output:
object Person {
  val name = monocle.macros.internal.Macro.mkLens[Person, String]("name");
  val age = monocle.macros.internal.Macro.mkLens[Person, Int]("age")
}

// input:
@Lenses case class Foo[A, B](q: Map[(A, B), Double], default: Double)

// output:
object Foo {
  def q[A, B] = monocle.macros.internal.Macro.mkLens[Foo[A, B], Map[(A, B), Double]]("q");
  def default[A, B] = monocle.macros.internal.Macro.mkLens[Foo[A, B], Double]("default")
}

### CODE REVIEW

**Comment by refried** [comment]
derp, I didn't think to cross-build before PR.

**Comment by refried** [comment]
Wait the error was this:
```[error] (monocle-core/*:update) sbt.ResolveException: download failed: org.scalaz#scalaz-core_2.11;7.1.1!scalaz-core_2.11.jar(bundle)```
That's not me. :angry:   How do you make Travis just re-run it again as-is?

**Comment by refried** [comment]
Closing and re-opening PR to trigger rebuild.

**Comment by refried** [comment]
Ok, that's better.

File: example/src/test/scala/monocle/LensDefExample.scala [hunk 1]
```diff
@@ -0,0 +1,41 @@
+package monocle
+
+import monocle.macros.{GenLens, Lenses}
+import monocle.syntax._
+import org.specs2.execute.AnyValueAsResult
+import org.specs2.scalaz.Spec
+import shapeless.test.illTyped
+
+class LensDefExample extends Spec {
```
julien-truffaut: would you mind to merge this example with LensExample.scala, it would be nice to have all the possible ways to create `Lens` in the same place

macro/src/main/scala/monocle/macros/Lenses.scala
```diff
@@ -20,10 +20,14 @@ private[macros] object LensesImpl extends MacrosCompatibility {
       case _ => ""
     }
 
-    def lenses(tpname: TypeName, paramss: List[List[ValDef]]): List[Tree] = {
+    def lenses(tpname: TypeName, tparams: List[TypeDef], paramss: List[List[ValDef]]): List[Tree] = {
       paramss.head map { param =>
         val lensName = createTermName(c)(prefix + param.name.decodedName)
-        q"""val $lensName = monocle.macros.internal.Macro.mkLens[$tpname, ${param.tpt}](${param.name.toString})"""
+        if (tparams.isEmpty)
```
julien-truffaut: that's nice it doesn't need a complete rewrite :)

julien-truffaut:
really cool, thanks @refried. Do you know if it is possible to write a test to check if the `Lens` generated for case classes without type parameter is `val` and not a `def`?

example/src/test/scala/monocle/LensDefExample.scala
refried: Good idea, yes.

refried:
I'm not sure how to do that, but I'll take a quick look.

refried:
@julien-truffaut I wrote a test that checks `val` vs `def` but the method I'm using was renamed between scala 2.10 and 2.11.  So we can use the deprecated 2.10 syntax, or not include the test, or do something uglier separate tests for 2.10 and 2.11.  What do you want to do?

julien-truffaut:
let's use the deprecated name for the moment, we can either refactor later or use 2.11 name when we stop supporting 2.10.

refried:
It looks like a big change due to moved lines for improved formatting.  If you want to re-rearrange them though, I support you :-)

julien-truffaut:
Great work, Thank you

refried:
Glad I could help.

dwijnand:
Nice one @refried

PR #191, opened 2015-03-06
Fix #190 use public PLens constructor in macro

### LINKED ISSUE #190: Macro does not work outside of monocle package
```scala
package foo
import monocle.macro.GenLens
case class Foo(n: Int)
object Foo {
  val lens = GenLens[Foo](_.n)
}
```

causes:

```
Error:(23, 18) constructor PLens in class PLens cannot be accessed in <$anon: monocle.PLens[foo.Foo,foo.Foo,Int,Int]>
  GenLens[Custom](_.value)
                 ^
```

I suppose the problem is that macro use the call site to determine package visibility. Since `Lens` constructor is private to monocle, it cannot generate lenses outside of monocle package :/

Thanks for @refried for finding this.

refried:


julien-truffaut:
This is one solution but it will have a big impact on performance (see [benchmarks](https://github.com/julien-truffaut/Monocle/wiki/Lens-Benchmark#use-random-input-from-jmh-and-measure-in-%CE%BCsops-28022015)).

We could also decide to make the constructor public but I would rather not expose implementation details. Before making a decision, I would like to check if we could access private constructor from macros or use definition site instead of call site regarding package visibility. We might need to invoke @xeno-by jin.

refried:
Right, I wondered about that, but I assumed that if performance was an issue -- well, I figured there must be some public constructor that allows for good performance, right?

If there is currently no high-performing version that works outside of the monocle package (i.e. the existing macros are un-callable), I'd encourage you to merge this for now and even release a binary!, and then certainly open a ticket to improve the performance.

refried:
Monocle operations are already so much faster than other libraries.  It's great if we can get an extra 20% improvement by inlining the code, but not if you can't call it :disappointed:   Let me know if I've misunderstood!

julien-truffaut:
@refried just to be clear, did you face this issue with 1.0.1 or with snapshot? I think it was only introduced in master, if that's correct it is worth to spend some time to look for a workaround.

I agree with you that the performance increase is too big to miss it but I believe it is only the case for case classes (because they have a built-in accessor and copy function).

xeno-by:
@julien-truffaut No, unfortunately you can't defy visibility rules with macros. In some situations this might be useful, but our macro system doesn't allow that.

refried:
@julien-truffaut I'm not sure, I only got serious about Monocle once the initial Intellij patch was put into place.  I did successfully use Lenser in the past though.  Is that an answer to the right question?

Wrt workaround, could you describe the ideal solution? From there, if it's not impossible, we can try to find it :-)

refried:
(I'm having trouble imagining it.  It seems like either the fast code has to be public, or we can only use slow code.)

refried:
Ok so in 1.0.1, the macro uses `Lens.apply`, similar to the change here in #191, which uses `PLens.apply`.  Which version was the one you benchmarked?

julien-truffaut:
@xeno-by thanks, that's probably a good thing

@refried the benchmarks are against master but we should update the bench module to depend on a specific version of monocle, anyway we can always use snapshots. In that particular case, I benchmarked 1.0.1 before releasing it and macros were as fast as apply method which is why I changed the macro to use private constructor in master.

PR #192, opened 2015-03-06
Fix #190 Allow users to create new PLens

### LINKED ISSUE #190: Macro does not work outside of monocle package

refried:
As compared to creating lenses with `PLens.apply`, creating lenses with `new PLens { ... }` results in lenses which are -18% to 78% faster in various [benchmarks](https://github.com/julien-truffaut/Monocle/wiki/Lens-Benchmark).

We had hoped that we could use macros to inline these new PLenses in user code, but unfortunately package private access restrictions are enforced on the expanded macros.  

Since macros essentially insert Scala source, I don't think we can inline code that would violate JVM permissions. :disappointed:

### DIFF
--- a/core/src/main/scala/monocle/Lens.scala
+++ b/core/src/main/scala/monocle/Lens.scala
@@ -30,7 +30,7 @@ import scalaz.{Applicative, Choice, Functor, Maybe, Monoid, Split, \/}
  * @tparam A the target of a [[PLens]]
  * @tparam B the modified target of a [[PLens]]
  */
-abstract class PLens[S, T, A, B] private[monocle]{ self =>
+abstract class PLens[S, T, A, B] { self =>
 
   /** get the target of a [[PLens]] */
   def get(s: S): A
--- a/test/src/test/scala/user/Visbility.scala
+++ b/test/src/test/scala/user/Visbility.scala
@@ -1,0 +1,12 @@
+package user
+
+class Visbility {
+
+  /* Macro should work outside of monocle package */
+  import monocle.macros.GenLens
+  case class Foo(n: Int)
+  object Foo {
+    val lens = GenLens[Foo](_.n)
+  }
+}
+


julien-truffaut:
I think it is the right way to go. We discussed in gitter with @NightRa about using internal package to signal that using `PLens` constructor is not recommended but I am not sure how to implement it.

@refried could you make the same change for other optics? I am almost sure that `GenPrism` and `GenIso` will have the same private constructor issue.

NightRa:
@julien-truffaut We could for the time being write it in the docs.
Also:
http://stackoverflow.com/questions/15897014/scala-constructor-deprecation

refried:
@julien-truffaut Sure.  Do you want to try running the benchmarks on the #191 branch just for the heck of it first to see if it really still makes a difference though?  Sorry I'm not at my normal development machine for a while.

julien-truffaut:
I will try to update the benchmark to run on 1.0.1 by the end of the week but I am almost sure they were no difference between macro and apply method (since the former use the later)

refried:
I'll update the PR as soon as I can, to open up the other classes like you
suggested; I've been swamped though, sorry.

julien-truffaut:
@refried np take your time

julien-truffaut:
@refried I made all constructors public

refried:
@julien-truffaut Ok thanks!

PR #204, opened 2015-04-08
Enhance GenLens to support select chains i.e. _.a.b.c

cvogt:
hi from @xdotai :)


julien-truffaut:
@cvogt really nice, thank you. 

Could you tell me what error message do you get if an accessor along the chain is not a case class accessor?

cvogt:
Well something like `GenLens[A](_.b.c.d)` is simply expanded into `GenLens[A].b composeLens GenLens[B](_.c) composeLens GenLens[C](_.d)` (assuming `.b: B, .c:C`). You tell me what Monocle does when you try to use GenLens on a non-case class field :). I didn't check.

PR #210, opened 2015-05-17
Some micro-optim. in Optional/Lens apply.

jbgi:
* Removed unnecessary indirections via 
  getOrModify in Optional.
* Removed unnecessary indirections via
  set in Lens and Optional.


julien-truffaut:
good catch, thank you @jbgi

PR #219, opened 2015-06-22
GenIso for objects

adelbertc:
Many thanks to @retronym for helping me out with a Scala 2.10/2.11 issue regarding prefixes, see https://gist.github.com/retronym/4752d7c1dba6fb837ae5 (reproduced below)

```scala
% tail sandbox/{macro,client}.scala
==> sandbox/macro.scala <==
object Macro {
  def m[S]: () => S = macro impl[S]
 
  def impl[S : c.WeakTypeTag](c: Context): c.Expr[() => S] = {
    import c.universe._
    val sym = weakTypeOf[S].typeSymbol.asClass.module
 
    reify(() => c.Expr[S](Ident(sym)).splice)
  }
}
 
==> sandbox/client.scala <==
class Client {
  object O
  Macro.m[O.type]
}
% (export V=v2.11.6; scalac-hash $V sandbox/macro.scala && scalac-hash $V sandbox/client.scala)
warning: there was one deprecation warning; re-run with -deprecation for details
one warning found
 
% (export V=v2.10.4; scalac-hash $V sandbox/macro.scala && scalac-hash $V sandbox/client.scala)
sandbox/client.scala:3: error: type mismatch;
 found   : O.type
 required: Client.this.O.type
  Macro.m[O.type]
         ^
one error found
```

### DIFF
--- a/macro/src/main/scala-2.10/monocle.macros.internal/MacrosCompatibility.scala
+++ b/macro/src/main/scala-2.10/monocle.macros.internal/MacrosCompatibility.scala
@@ -1,5 +1,7 @@
 package monocle.macros.internal
 
+import scala.reflect.macros.TreeBuilder
+
 trait MacrosCompatibility {
   type Context = scala.reflect.macros.Context
 
@@ -26,4 +28,7 @@ trait MacrosCompatibility {
 
   def companionTpe(c: Context)(tpe: c.universe.Type): c.universe.Symbol =
     tpe.typeSymbol.companionSymbol
+
+  def makeAttributedQualifier(c: Context)(tree: TreeBuilder, tpe: c.universe.Type): c.universe.Tree =
+    tree.mkAttributedQualifier(tpe.asInstanceOf[tree.global.Type]).asInstanceOf[c.universe.Tree]
 }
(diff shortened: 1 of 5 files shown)


adelbertc:
Note: `TreeBuilder#gen.mkAttributedQualifier` is an internal API in Scala 2.10, in Scala 2.11.x+ there is a public version `internal.treeBuild.mkAttributedQualifier`

julien-truffaut:
cool, it means we can implement a `Prism` between an ADT and an object by composing two macros:
```scala
import monocle.macro._

sealed trait Foo
case object Bar extends Foo

val _bar: Prism[Foo, Unit] = GenPrism[Foo, Bar] composeIso GenIso.obj[Bar]
```

If I understood correctly the implementation generates a warning because of difference between 2.10 and 2.11. If that's correct, could you please define the specific method one for each scala version (2.10 and 2.11) in:
*   https://github.com/julien-truffaut/Monocle/blob/master/macro/src/main/scala-2.10/monocle.macros.internal/MacrosCompatibility.scala
*   https://github.com/julien-truffaut/Monocle/tree/master/macro/src/main/scala-2.11/monocle.macros.internal

Issue #222, opened 2015-06-22
GenIso between empty case classes and Unit


adelbertc:
```scala
case class Foo[A]()
case class Bar()

GenLens.empty[Foo, Unit]
GenLens.empty[Bar, Unit]
```

adelbertc:
Closed https://github.com/julien-truffaut/Monocle/pull/223

PR #223, opened 2015-06-24
GenIso for empty case classes

adelbertc:


adelbertc:
Closes #222

adelbertc:
Hm build seems to fail with an issue pulling 2.11.6 ?

julien-truffaut:
That's nice you managed to handle the object and empty case classes in the same method :)

what do you think about `apply0`, `one` or `single` instead of `unit`? In fact, `Unit` is by definition the type with a single value, so maybe that's the most accurate name.

PR #234, opened 2015-08-17
renamed getOrModify to matching

julien-truffaut:
I never liked the name `getOrModify` but I couldn't find the equivalent in haskell lens.

I completely missed it: https://hackage.haskell.org/package/lens-4.12.3/docs/src/Control-Lens-Prism.html#matching

### DIFF
--- a/core/src/main/scala/monocle/Iso.scala
+++ b/core/src/main/scala/monocle/Iso.scala
@@ -200,7 +200,7 @@ abstract class PIso[S, T, A, B] extends Serializable { self =>
   /** view a [[PIso]] as a [[POptional]] */
   @inline final def asOptional: POptional[S, T, A, B] =
     new POptional[S, T, A, B]{
-      def getOrModify(s: S): T \/ A =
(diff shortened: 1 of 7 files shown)


NightRa:
I personally don't think it's a good idea to break compatibility for this kind of name change.
Also, the name `matching` isn't any more intuitive than `getOrModify`.

julien-truffaut:
You're right I might have overlooked the argument of binary compatibility. 

The good thing is that I believe that `getOrModify` is rarely used, so I don't think it would be too harmful to rename it in the next non binary compatible release (1.2.0). Btw, I don't know if you noticed but I let `getOrModify` but deprecated it.

`matching` talks more to me that `getOrModify` but that's probably personal.

julien-truffaut:
@NightRa do you think the aliasing + deprecation is acceptable or we should rather revert this change?

julien-truffaut:
@NightRa I rolledback this change, I agree with you there is no point to make such a change now.

PR #236, opened 2015-08-26
add length method for Fold

aoiroaoino:
https://hackage.haskell.org/package/lens-4.12.3/docs/Control-Lens-Fold.html#v:lengthOf


julien-truffaut:
good idea :) could you also add it for `Traversal`?

aoiroaoino:
OK, I’ll get right on it:smile:

julien-truffaut:
great!

PR #250, opened 2015-09-23
remove kind projector (limited usage + intellij issue)

julien-truffaut:
This is a contentious PR.

Intellij 14 / 15 CE does not seem to understand kind projector any more. I believe it started since we migrated from `Build.scala` to `build.sbt` but I cannot tell for sure. Since it is blocking or at least annoying potential contributor (see https://groups.google.com/forum/#!topic/scala-monocle/2MEXvY4k5cE) I would rather remove `kind projector` for the moment.

### DIFF
--- a/core/src/main/scala/monocle/Traversal.scala
+++ b/core/src/main/scala/monocle/Traversal.scala
@@ -38,7 +38,7 @@ abstract class PTraversal[S, T, A, B] extends Serializable { self =>
 
   /** map each target to a Monoid and combine the results */
   @inline final def foldMap[M: Monoid](f: A => M)(s: S): M =
-    modifyF[Const[M, ?]](a => Const(f(a)))(s).getConst
+    modifyF[({type λ[α] = Const[M, α]})#λ](a => Const(f(a)))(s).getConst
 
   /** combine all targets using a target's Monoid */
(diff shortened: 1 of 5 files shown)


NightRa:
Kind Projector got full support in version 1.7 of the scala plugin, which is for **Intellij Idea 15 EAP**, so 14/15 CE doesn't have it yet.
:+1: for this as we don't use it much.

core/src/main/scala/monocle/Traversal.scala
```diff
@@ -38,7 +38,7 @@ abstract class PTraversal[S, T, A, B] extends Serializable { self =>
 
   /** map each target to a Monoid and combine the results */
   @inline final def foldMap[M: Monoid](f: A => M)(s: S): M =
-    modifyF[Const[M, ?]](a => Const(f(a)))(s).getConst
+    modifyF[({type λ[α] = Const[M, α]})#λ](a => Const(f(a)))(s).getConst
```
NightRa: What do you think about giving the impl. with kind projector as a comment for readability?
This seems like the only justified use of kind projector which improves understanding.
But it would make the formatting of the file less clean..

julien-truffaut: why not, you're right is the most interesting usage of kind projector. Adding a comment could help reader

PR #254, opened 2015-09-25
attempt to fix auto publish

julien-truffaut:
I regenerated sonatype credential, who knows if it will help ...


julien-truffaut:
since there is no way to check if it works before hand, I am going to merge ... and cross my fingers ^^

PR #255, opened 2015-09-26
Rename sum to choice and product to split

julien-truffaut:
The idea is to have consistent method name with typeclass.


aoiroaoino:
LGTM:+1:

NightRa:
:+1: 
Good, now it's more straightforward to see this is just infix syntax for the class's function.

PR #256, opened 2015-09-30
Add documentation

justjoheinz:
Fix errors and add some smaller hints, like using modifyF with a Future (common use case)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: scala
 scala:
 - 2.11.7
-- 2.10.5
+- 2.10.6
 jdk:
 - oraclejdk7
 notifications:
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,9 @@ notifications:
     on_start: false
   email:
   - truffaut.julien@gmail.com
-script: sbt ++$TRAVIS_SCALA_VERSION test
+script:
+  - sbt ++$TRAVIS_SCALA_VERSION test
+  - sbt ++$TRAVIS_SCALA_VERSION tut
 env:
   global:
   - secure: JjrFp+1hGEwifDp5vh9BZ4vlZKe/zg/KbGg57PuCtsEgOvpUhvu0Fl6b3JImmF8GKZNjhk2EcgWAA8wyROIORVYNVaroOTHf/XoLvrHhgskGpx042O3gDuiufAaItUGN6NWXB/IjafP69Y1OWDocOsDWRJuL7f/9uom9aKdpamM=
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,7 @@ script:
   - sbt ++$TRAVIS_SCALA_VERSION tut
 env:
   global:
-  - secure: JjrFp+1hGEwifDp5vh9BZ4vlZKe/zg/KbGg57PuCtsEgOvpUhvu0Fl6b3JImmF8GKZNjhk2EcgWAA8wyROIORVYNVaroOTHf/XoLvrHhgskGpx042O3gDuiufAaItUGN6NWXB/IjafP69Y1OWDocOsDWRJuL7f/9uom9aKdpamM=
-  - secure: bNUYybjKmsptOCsOsgEX29WKPwSZQP9N9fKAyyc/zIIq4tyAr5mJ/v3YRdEB7n2F7cnpNJQqUZ5llQdNGeFUNb2YycbWzTbA2zQmuoI3pawaUwDii57yOMd+E0I21eiok6eXaMlm9RzOBJ8iRwO9uL7QnUPuNqXbUwNT2zkspS4=
-
+  - secure: lJxaR+9VM96s7rfaigL3K2FbcP3XiZ+AerbgOBOpYi2anRkMPQ91h1g60HYTwEDGJIKybyNWFuXZM8hNtQMCxzxgu7ev1SyRr5Otfzg5ehGqB5UWbSAgpXlqhR5dAonwmqyEEDF2OK+k1fa0N6mIwpQeuUCwrt68kTp4f0K9jMQ=
+  - secure: NZQn8lB04I+FO3Om680CXWP29scS/zRhITbLO/1sM3dnuG8ygn0fzlFMQhgCO3fCjnclThAi+6MN7UnyV8BwDbFhCDe627+06px5fHUw2TPdzhz/BwngAW9pRsx6m0b8eBqQmt8Q6mlgeYOPemsFOeqQTKZulpIhSYnO1R1Nwzc=
 after_success:
-  - "[[ $TRAVIS_BRANCH == \"master\" ]] && { sbt publish; };"
+- '[[ $TRAVIS_BRANCH == "master" ]] && { sbt publish; };'
--- a/docs/src/main/tut/iso.md
+++ b/docs/src/main/tut/iso.md
@@ -1,0 +1,34 @@
+---
+layout: default
+title:  "Iso"
+section: "optics"
+scaladoc: "http://julien-truffaut.github.io/Monocle/api/#monocle.PLens"
+source: "https://github.com/julien-truffaut/Monocle/blob/master/example/src/main/scala/monocle/example/LensExample.scala"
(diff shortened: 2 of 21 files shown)


justjoheinz:
I plan to add some more documentation. Please let me know if you prefer one commit or smaller ones. I prefer smaller ones, as I think less documentation is better than none at all. Also please review - part of the excercise for me is to get a better understanding of optics.

docs/src/main/tut/iso.md
```diff
@@ -0,0 +1,34 @@
...
+title:  "Iso"
+section: "optics"
+scaladoc: "http://julien-truffaut.github.io/Monocle/api/#monocle.PLens"
+source: "https://github.com/julien-truffaut/Monocle/blob/master/example/src/main/scala/monocle/example/LensExample.scala"
+pageSource: "https://raw.githubusercontent.com/julien-truffaut/Monocle/master/docs/src/main/tut/iso.md"
+---
+
+# Iso
+
+An `Iso` is an Optic which converts elements of type `A` into elements of type
```
julien-truffaut: We generally use type `S` and `A`. It doesn't really matter but I suppose it is better to be consistent.

docs/src/main/tut/iso.md
```diff
@@ -0,0 +1,34 @@
...
+`B` without loss.
+
+Consider these two case classes:
+
+```tut:silent
+case class Person(name: String, firstName : String)
```
julien-truffaut: This example might be more illustrative if the fields in `Person` are of different type e.g. `String` and `Int`

docs/src/main/tut/iso.md
```diff
@@ -0,0 +1,34 @@
...
+case class Pers(n: String, f: String)
+```
+
+In order to create an `Iso` between `Person` and `Pers` we need to supply two total functions
+
+* get : Person => Pers
+* reverseGet : Pers => Person
+
+```tut:silent
+import monocle._
```
julien-truffaut: maybe we could use a more precise import, e.g. `monocle.Iso`

docs/src/main/tut/iso.md
```diff
@@ -0,0 +1,34 @@
...
+val personPrism = Iso[Person, Pers]((p: Person) => Pers(p.name, p.firstName))((p: Pers) => Person(p.n, p.f))
```
julien-truffaut: if you want to make this example shorter you can either use:
`Iso[Person, Pers](p => Pers(p.name, p.firstName))(p => Person(p.n, p.f))`
or
Iso((p: Person) => Pers(p.name, p.firstName))((p: Pers) => Person(p.n, p.f))

docs/src/main/tut/iso.md
```diff
@@ -0,0 +1,34 @@
...
+section: "optics"
+scaladoc: "http://julien-truffaut.github.io/Monocle/api/#monocle.PLens"
+source: "https://github.com/julien-truffaut/Monocle/blob/master/example/src/main/scala/monocle/example/LensExample.scala"
+pageSource: "https://raw.githubusercontent.com/julien-truffaut/Monocle/master/docs/src/main/tut/iso.md"
+---
+
+# Iso
+
+An `Iso` is an Optic which converts elements of type `A` into elements of type
+`B` without loss.
```
julien-truffaut: later we might want to mention that `Iso` comes from Isomorphism and point to wikipedia or other resources

docs/src/main/tut/iso.md
```diff
@@ -0,0 +1,34 @@
...
+* reverseGet : Pers => Person
+
+```tut:silent
+import monocle._
+val personPrism = Iso[Person, Pers]((p: Person) => Pers(p.name, p.firstName))((p: Pers) => Person(p.n, p.f))
+```
+
+and thereby create a lossless conversion between these two types. Other usages include for example the transformation of different types of physical units.
+
+It is important to understand that the transformation between type `A` and `B` works for any type `A` and `B` and that the functions `get` and `reverseGet` are true inverses to each other.
```
julien-truffaut: :+1: to mention that `get` and `reverseGet` are inverse of each other, we could also give the law that comes from this. However, I find the first sentence ("It is important") a bit confusing.

docs/src/main/tut/lens.md
```diff
@@ -50,15 +50,32 @@ We can push push the idea even further, with `modifyF` we can update the target
...
+
 import scalaz.std.list._ // to get Functor[List] instance
 ```
+
 ```tut
 _streetNumber.modifyF(neighbors)(address)
 _streetNumber.modifyF(neighbors)(Address(135, "High Street"))
 ```
 
+This would work with any kind of `Functor` and is especially useful in conjunction with asynchronous APIs, where one has the task to update a deeply nested structure (see Lens Composition) with the result of an asynchronous computation:
```
julien-truffaut: I would say any `Functor` or any type `F` that has an instance of `Functor`

docs/src/main/tut/iso.md
justjoheinz: In documentation I like to see the types even if they can be inferred as I believe it helps to connect the dots as I cannot get them tooltipped from the screen.

docs/src/main/tut/iso.md
justjoheinz: ok

docs/src/main/tut/iso.md
justjoheinz: ok

docs/src/site/optics.md
```diff
@@ -6,5 +6,12 @@ section: "optics"
 
 # Optics
 
+## Convert between two types
```
julien-truffaut: I am not sure we should make that kind of categorisation here. Maybe it would be a good place to put the class diagram that shows relationship between optics.

julien-truffaut:
Thanks @justjoheinz it is very helpful! I made many small comments, you don't have to take care of all of them, other people including yourself might disagree. Also I am not a native speaker, so I am definitevely not the best person to correct formulation.

docs/src/main/tut/iso.md
julien-truffaut: it makes sense, maybe we can show a shorter version in other example or as a note.

julien-truffaut: also the name of the `Iso` contains `Prism`

justjoheinz:
I just incorporated some changes, and missed some you were obviously just making. I also added a "tut" target to travis.yml - the docs were already containing errors, and I hope we agree, that this should not happen.

.travis.yml
```diff
@@ -13,7 +13,9 @@ notifications:
     on_start: false
   email:
   - truffaut.julien@gmail.com
-script: sbt ++$TRAVIS_SCALA_VERSION test
+script:
```
julien-truffaut: great I didn't notice

docs/src/site/optics.md
justjoheinz: I would like to rework that diagram eventually. It confused me totally when I tried to learn about monocle, because I constantly thought it was talking about inheritance and it took me awhile to realize that it is talking about transformations (I am UML damaged to certain degree). Thus I never understood why my lenses did not compose.

julien-truffaut: At some point (0.5 or 0.6), optics were actually inheriting from each other. We removed it when we started to encode them as case class, nowadays I am not sure if we should use or not inheritance. I

jbgi: Having access to the Prism (weaker) api in Iso would be confusing, better not use inheritance.

julien-truffaut:
@justjoheinz if you can rebase this PR, I will be happy to merge it.

julien-truffaut:
It is weird, it seems that they are still some conflict but we haven't merged anything in the meantime.

justjoheinz:
I think I mucked it up, when rebasing, will give it another try tommorow, had been a busy day.

julien-truffaut:
np, please tell me if I can help :)

julien-truffaut:
@justjoheinz I propose that we close this PR until you or anyone else get sometimes to work on it.

justjoheinz:
Ok, maybe the easiest for me would be to close this, and just reopen a new PR based on the current state. Sorry - have been away and busy.

justjoheinz:
I just opened another branch created from a clean slate.

PR #258, opened 2015-10-03
modify to use NonEmptyList#zipWithIndex

aoiroaoino:
https://github.com/scalaz/scalaz/blob/v7.1.4/core/src/main/scala/scalaz/NonEmptyList.scala#L117


julien-truffaut:
:+1:

PR #259, opened 2015-10-06
Fix typo in lens.md

1ambda:
`Lens[S, A].apply` requires `A => S => S` as the second argument

So, `(p => a => p.copy...)` should be `(a => p => p.copy...)`

### DIFF
--- a/docs/src/main/tut/lens.md
+++ b/docs/src/main/tut/lens.md
@@ -67,7 +67,7 @@ val john = Person("John", 20, address)
 ```
 
 ```scala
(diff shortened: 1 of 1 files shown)


julien-truffaut:
good catch! In fact, we should use `tut` for that kind of example but I remember getting some weird error related to scala console.

:+1:

justjoheinz:
Yes, sometime `tut` /console  behaves really weird. I am just struggling with that in terms of the tutorial pages which I try to write.

PR #262, opened 2015-10-09
Add HttpRequestExample.scala

1ambda:
I'v implemented the `HttpRequest` example used in [Beyond scala Lens](http://www.slideshare.net/JulienTruffaut/beyond-scala-lens)

This example shows 

- how to use `GenPrism` to create `Prism` with case object. I found #219 but it seems outdated. (`GenIso.obj` instead of `GenIso.unit`) 

```scala
sealed trait HttpMethod
case object GET   extends HttpMethod
case object POST  extends HttpMethod

val get: Prism[HttpMethod, Unit] = GenPrism[HttpMethod, GET.type] composeIso GenIso.unit[GET.type]
val post = GenPrism[HttpMethod, POST.type] composeIso GenIso.unit[POST.type]
```

- how to use `at` and `index` to achieve the same result, and how they are different
- `Gen*` macro, `filterIndex` usage

### DIFF
--- a/example/src/test/scala/monocle/HttpRequestExample.scala
+++ b/example/src/test/scala/monocle/HttpRequestExample.scala
@@ -75,7 +75,6 @@ class HttpRequestExample extends MonocleSuite {
       composeTraversal filterIndex { h: String => h.contains("timeout") }
       composePrism stringToInt).modify(_ * 2)(r1)
 
-    println(r)
     r.headers.get("socket_timeout") shouldBe Some("40")
     r.headers.get("connection_timeout") shouldBe Some("20")
   }
--- a/example/src/test/scala/monocle/HttpRequestExample.scala
+++ b/example/src/test/scala/monocle/HttpRequestExample.scala
@@ -1,0 +1,91 @@
+package monocle
+
+import monocle.macros.{GenLens, GenIso, GenPrism}
+
+/**
+ * Show how could we use Monocle to handle custom case classes, objects
+ */
+class HttpRequestExample extends MonocleSuite {
+  import HttpRequestExample._
+
+  val r1 = HttpRequest(
+    GET,
+    URI("localhost", 8080, "/ping", Map("hop" -> "5")),
+    Map("socket_timeout" -> "20", "connection_timeout" -> "10"),
+    "")
+
+  val r2 = HttpRequest(
+    POST,
+    URI("gooogle.com", 443, "/search", Map("keyword" -> "monocle")),
+    Map.empty,
+    "")
+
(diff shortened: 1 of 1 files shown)


example/src/test/scala/monocle/HttpRequestExample.scala
```diff
@@ -0,0 +1,91 @@
...
+      composeLens query
+      composeOptional index("hop")
+      composePrism stringToInt).modify(_ + 10)(r1)
+
+    r.uri.query.get("hop") shouldBe Some("15")
+  }
+
+  test("query using at") {
+    /**
+     *  `at` returns Lens[S, Option[A]] while `index` returns Optional[S, A]
```
julien-truffaut: here you could use `composeOption index("hop")`, I am not sure it is what you meant

julien-truffaut:
good idea! In a second step, I think we could move this example to the website and use `tut` to show the results

example/src/test/scala/monocle/HttpRequestExample.scala
```diff
@@ -0,0 +1,91 @@
...
+    val r = (headers composeLens at("Content-Type")).set(Some("text/plain; utf-8"))(r2)
+    r.headers.get("Content-Type") shouldBe Some("text/plain; utf-8")
+  }
+
+  test("headers with filterIndex") {
+    val r = (headers
+      composeTraversal filterIndex { h: String => h.contains("timeout") }
+      composePrism stringToInt).modify(_ * 2)(r1)
+
+    println(r)
```
aoiroaoino: debug code ?

1ambda:
@aoiroaoino I removed the debug code. Thanks :) 

@julien-truffaut The `hop` means [Network Hop (distance between nodes)](http://www.wikiwand.com/en/Hop_(networking)). Since the `r1` is a simulation of network `Ping` request , I used `hop` as a query string parameter. 

```scala
// localhost:8080/ping?hop=5

val r1 = HttpRequest(
  GET,
  URI("localhost", 8080, "/ping", Map("hop" -> "5")),
  Map("socket_timeout" -> "20", "connection_timeout" -> "10"),
  "")
```

julien-truffaut:
No it makes sense to show two different options, maybe it will be clearer if we would say something like

`at("hop") composePrism some` is equivalent to `index("hop")` because `at` returns `Lens[S, Option[A]]` while `index` returns `Optional[S, A]`

julien-truffaut:
Thanks @1ambda for this contribution! I will merge it and I or someone else can try to turn this example into some `tut` examples later on.

1ambda:
Thanks :)

PR #264, opened 2015-10-10
fix #245 add fromIso constructor for typeclass

### LINKED ISSUE #245: Create typeclass helper using `Iso`
```scala 
def eachFromIso[S, A, B](iso: Iso[S, A])(implicit ev: Each[A, B]): Each[S, B] = new Each[S, B] {
  def each: Traversal[S, B] = iso composeTraversal ev.each
}
```

julien-truffaut:


julien-truffaut:
@aoiroaoino @NightRa if you get some times, would you mind to review this PR.

aoiroaoino:
oops, sorry..
I will check on it.

julien-truffaut:
thanks @aoiroaoino, there is nothing urgent but it is really useful to have a second pair of eyes looking at it.

test/src/test/scala/monocle/function/AtSpec.scala
```diff
@@ -0,0 +1,12 @@
+package monocle.function
+
+import monocle.MonocleSuite
+import monocle.law.discipline.function.AtTests
+
+class AtSpec extends MonocleSuite {
+
+  implicit def mmapAt[K, V]: At[MMap[K, V], K, V] = At.fromIso(MMap.toMap)
+
+  checkAll("fromIso", AtTests.defaultIntIndex[MMap[Int, String], String])
```
NightRa: What's the point of defaultIntIndex? Does it just check for 2?

test/src/test/scala/monocle/TestInstances.scala
```diff
@@ -27,7 +27,7 @@ trait TestInstances {
   implicit val byteEqual    = Equal.equalA[Byte]
   implicit val shortEqual   = Equal.equalA[Short]
   implicit val charEqual    = Equal.equalA[Char]
-  implicit val intEqual     = Equal.equalA[Int]
+//  implicit val intEqual     = Equal.equalA[Int]
```
NightRa: Any reason for this?

core/src/main/scala/monocle/function/Snoc1.scala
```diff
@@ -20,7 +20,12 @@ trait Snoc1[S, I, L] extends Serializable {
   def last: Lens[S, L] = snoc1 composeLens second
 }
 
-object Snoc1 extends Snoc1Functions
+object Snoc1 extends Snoc1Functions {
+  def fromIso[S, A, I, L](iso: Iso[S, A])(implicit ev: Snoc1[A, I, L]): Snoc1[S, I, L] = new Snoc1[S, I, L] {
+    override def snoc1: Iso[S, (I, L)] =
+      iso composeIso ev.snoc1
```
NightRa: Add documentation?
This won't compose from the other side like in Snoc because we have freedom in the tail; for example, NEL[A] = (A,List[A])

test/src/test/scala/monocle/TestInstances.scala
julien-truffaut: nop, I commented it when I was debugging something unrelated. Well spotted

core/src/main/scala/monocle/function/Field1.scala
```diff
@@ -1,21 +1,22 @@
...
   /** Creates a Lens from S to it is first element */
   def first: Lens[S, A]
-
 }
 
-object Field1 extends Field1Functions
+object Field1 extends Field1Functions {
+  def fromIso[S, A, B](iso: Iso[S, A])(implicit ev: Field1[A, B]): Field1[S, B] = new Field1[S, B] {
+    override def first: Lens[S, B] =
+      iso composeLens  ev.first
```
NightRa: 2 spaces. How did I manage spot it? XD

test/src/test/scala/monocle/function/AtSpec.scala
julien-truffaut: ideally it should generate indexes and check `LensLaw` for them but I didn't know how to integrate that. So the temporary and ugly solution I found was to use the laws for a fixed index ...

NightRa:
:+1:

core/src/main/scala/monocle/function/Field1.scala
julien-truffaut: you must be half robot ;)

core/src/main/scala/monocle/function/Snoc1.scala
julien-truffaut: I am not sure to understand what do you mean by "it won't compose"?

NightRa:
Should also document why we chose to put the Iso in this direction and not the other (although they are equivalent as this is an iso and we can reverse it) - 
(A -> B), F[B] -> F[A]
This comes from the perspective of "I want an F[A] and this is how I'll turn it into something I already have" vs
F[B], (B -> A) -> F[A]
"This is what I have, and here's how I can use it to get the F[A] I want"

core/src/main/scala/monocle/function/Snoc1.scala
NightRa: Applying the Iso to the tail too, like in Snoc & Cons.

julien-truffaut:
very good point, I guess one thing that is not documented is that in our type classes is that `A` is supposed to be unique given an `S`, e.g. `Cons[List[A], A]`. I think in haskell we can express such type constraints but as far as I know we cannot in scala. Anyway it should be documented

core/src/main/scala/monocle/function/Field1.scala
```diff
@@ -1,21 +1,22 @@
 package monocle.function
 
-import monocle.Lens
+import monocle._
```
aoiroaoino: I think, we should write explicit a import name.
`import monocle.{Lens, Iso}`

julien-truffaut: right, copy pasting in intellij does this. thanks

julien-truffaut:
@NightRa I added some doc, WDYT?

julien-truffaut:
Thank you guys for the detail review!

PR #273, opened 2015-10-17
left & right

julien-truffaut:
I don't think we can write a lawful `left` or `right` for any `Lens`, `Optional` or above. The main issue comes from the definition of `set`.


aoiroaoino:
looks good :+1:

However, I don't have the necessary knowledge to judge the concepts.
but, I'm going to trust you on this Pull Request.

julien-truffaut:
@aoiroaoino sorry I didn't see your comment. Thanks for the review, if we find out a way to generalise `left` and `right` to other optics, we can always add it later.

PR #274, opened 2015-10-18
Use refined-scalacheck for Arbitrary instances

fthomas:
... of refined types like `Minute` and `Hour`


aoiroaoino:
LGTM :+1:

@julien-truffaut WDYT?

julien-truffaut:
Very nice :+1:

Issue #248, opened 2015-09-23
Use macro-compat instead of adhoc trait


julien-truffaut:
see https://github.com/milessabin/macro-compat

aoiroaoino:
I have tried this issue.
https://github.com/julien-truffaut/Monocle/compare/master...aoiroaoino:topic/use_macro-compat

When v2.11.7, we passed all tests.:smile: 
But, I get compile errors in v2.10.6 ...

```sbt
> compile
[info] Compiling 6 Scala sources to /Users/aoino/git/Monocle/macro/target/scala-2.10/classes...
[error] /Users/aoino/git/Monocle/macro/src/main/scala/monocle/macros/GenIso.scala:35: value companion is not a member of Stub.this.c.universe.Symbol
[error]     val sTpeSym = sTpe.typeSymbol.companion
[error]                                   ^
[error] /Users/aoino/git/Monocle/macro/src/main/scala/monocle/macros/GenIso.scala:78: value companion is not a member of Stub.this.c.universe.Symbol
[error]           val sTpeSym = sTpe.typeSymbol.companion
[error]                                         ^
[error] /Users/aoino/git/Monocle/macro/src/main/scala/monocle/macros/GenPrism.scala:19: value companion is not a member of Stub.this.c.universe.Symbol
[error]     val sTpeSym = sTpe.typeSymbol.companion
[error]                                   ^
[error] /Users/aoino/git/Monocle/macro/src/main/scala/monocle/macros/internal/Macro.scala:61: value untypecheck is not a member of scala.reflect.macros.Context
[error]     val strFieldName = c.eval(c.Expr[String](c.untypecheck(fieldName.tree.duplicate)))
[error]                                                ^
[error] four errors found
[error] (macros/compile:compileIncremental) Compilation failed
[error] Total time: 4 s, completed 2015/10/06 20:25:32
```

maybe, `macro-compat` is not support compatible method following yet.
- Symbol#companion
- c.untypecheck

(However, all tests passed when I added method support in local.)
https://github.com/aoiroaoino/macro-compat/commit/35923555d9d4355ecc053124537bd303974c84f1

How do you think?

julien-truffaut:
nice! I don't know much about `macro-compat`, maybe it is worth to open a ticket or ask in the gitter channel https://gitter.im/milessabin/macro-compat

aoiroaoino:
Ok, I ask in the gitter.
Thanks! :laughing:

[...]

PR #276, opened 2015-10-21
Use macro-compat

aoiroaoino:
ref. #248

@milessabin @philwills @julien-truffaut many thanks! :smile:


macro/src/main/scala/monocle/macros/GenIso.scala
```diff
@@ -59,21 +60,22 @@ private object GenIsoImpl extends MacrosCompatibility {
...
 
-  def genIso_unit_impl[S: c.WeakTypeTag](c: Context): c.Expr[Iso[S, Unit]] = {
+  def genIso_unit_impl[S: c.WeakTypeTag]: c.Expr[Iso[S, Unit]] = {
     import c.universe._
 
     val sTpe = weakTypeOf[S]
 
     if (sTpe.typeSymbol.isModuleClass) {
       val table = c.universe.asInstanceOf[SymbolTable]
-      val obj = makeAttributedQualifier(c)(table.gen, sTpe)
```
milessabin: Would be worth adding `Context#internal.gen.mkAttributedQualifier` to macro-compat too.

julien-truffaut:
Great!

macro/src/main/scala/monocle/macros/GenIso.scala
aoiroaoino: ok, I will send a pull request :+1:

julien-truffaut:
:+1:

PR #283, opened 2015-11-03
Fix #114: generate polymorphic lenses

### LINKED ISSUE #114: Extend macros to generate polymorphic Lenses
Lenses and Lenser macro currently generate `SimpleLens`. It would be good if we could also generate `Lens` when the case class has a type parameter for an accessor, e.g.

```scala
case class Example[A](value: A, s: String, i: Int)
object Example {
  val lenser = Lenser[Example]
  val _s: Lens[Example, String] = lenser(_.s) // already exist
  def _value[A, B]: Lens[Example[A], Some[B], A, B] = lenser.heterogenousLens(_.value)
}
```

exlevan:
Here, `@Lenses` is changed to generate polymorphic lenses for case classes with type parameters. Generated lenses are polymorphic in types that are used by corresponding field (and only by it) and types that aren't used by any field. For example:
```
case class Foo[A, B, C](x: (A, B), y: B => Double)
```
will have the following lenses generated:
```
def x[A, B, C, A$, C$] = new PLens[Foo[A, B, C], Foo[A$, B, C$], (A, B), (A$, B)]{ … }
def y[A, B, C, C$] = new PLens[Foo[A, B, C], Foo[A, B, C$], B => Double, B => Double]{ … }
```

### DIFF
--- a/example/src/test/scala/monocle/LensExample.scala
+++ b/example/src/test/scala/monocle/LensExample.scala
@@ -94,4 +94,8 @@ class LensPolyExample extends MonocleSuite {
     Foo.q.modify((_: Map[(Int,Symbol),Double]).updated((0,'Buy), -2.0))(candyTrade) shouldEqual changedTrade
   }
 
+  test("@Lenses generates polymorphic lenses") {
+    val changedTrade = candyTrade.copy(q = candyTrade.q.map{case (x, y) => (x.swap, y)})
+    Foo.q.modify((_: Map[(Int, Symbol), Double]).map{case (x, y) => (x.swap, y)})(candyTrade) shouldEqual changedTrade
+  }
 }
--- a/example/src/test/scala/monocle/LensExample.scala
+++ b/example/src/test/scala/monocle/LensExample.scala
@@ -1,6 +1,6 @@
 package monocle
 
-import monocle.macros.{GenLens, Lenses}
+import monocle.macros.{GenLens, Lenses, PLenses}
 import shapeless.test.illTyped
 
 class LensMonoExample extends MonocleSuite {
@@ -69,7 +69,7 @@ class LensMonoExample extends MonocleSuite {
 
 class LensPolyExample extends MonocleSuite {
 
-  @Lenses case class Foo[A,B](q: Map[(A,B),Double], default: Double)
+  @PLenses case class Foo[A,B](q: Map[(A,B),Double], default: Double)
 
   object Manual { // Lens created manually (i.e. without macro)
   def q[A,B] = Lens((_: Foo[A,B]).q)(q => f => f.copy(q = q))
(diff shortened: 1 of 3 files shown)


julien-truffaut:
Awesome! So to be sure I understand the consequences:
```scala
@Lenses case class Foo[A, B, C](x: (A, B), y: B => Double)

// now 
object Foo {
  def x[A, B, C]: Lens[Foo[A, B, C], (A, B)] = ...
  def y[A, B, C]: Lens[Foo[A, B, C], B => Double] = ...
}

// after 
object Foo {
  def x[A, B, C, A$, B$]: PLens[Foo[A, B, C], Foo[A$, B$, C], (A, B), (A$, B$)] = ...
  def y[A, B, C, B$]: PLens[Foo[A, B, C], Foo[A, B$, C], B => Double, B$ => Double] = ...
}
```

If it is correct, it means that users who currently use `@Lenses` for case class with type parameters will probably need to add some type annotations (since type inference is likely to fail). I wondering if it is worth to keep existing behaviour and prefix the `PLens` you added with `p` or something else e.g.  

```scala
object Foo {
  def x[A, B, C]: Lens[Foo[A, B, C], (A, B)] = ...
  def y[A, B, C]: Lens[Foo[A, B, C], B => Double] = ...
  def px[A, B, C, A$, B$]: PLens[Foo[A, B, C], Foo[A$, B$, C], (A, B), (A$, B$)] = ...
  def py[A, B, C, B$]: PLens[Foo[A, B, C], Foo[A, B$, C], B => Double, B$ => Double] = ...
}
```

exlevan:
In your example, type param `B` can't be polymorphic, because it's shared between `x` and `y`, so it's impossible to change one field without changing the other. On the other hand, `C` is phantom type, so it's always is safe to change. So types will be like this:
```
// after 
object Foo {
  def x[A, B, C, A$, C$]: PLens[Foo[A, B, C], Foo[A$, B, C$], (A, B), (A$, B)] = ...
  def y[A, B, C, C$]: PLens[Foo[A, B, C], Foo[A, B, C$], B => Double, B => Double] = ...
}
```
Thinking about compatibility issue, another solution would be to have `@PLenses` annotation to make polymorphic lenses creation explicit and keep `@Lenses` unchanged. This way there will be no need to duplicate lens definitions and inventing name convention to differentiate between them.

julien-truffaut:
You're right shared type parameter cannot be shared. In that case, shall we generate a `Lens` instead of a `PLens`?

Good idea to create a separate macro annotation for `PLens`, it will make code simpler and maintain backward compatibility.

exlevan:
> shall we generate a `Lens` instead of a `PLens`?

Not in this case, since some type params are still can be changed (`A` and `C` for `x`, `C` for `y`). If there are no type params to change, then the generated `PLens[S, T, A, B]` will have `S =:= T` and `A =:=B`, so it will be naturally be a `Lens`.

julien-truffaut:
right, sorry I am a bit slow today. I didn't notice that `C` was a phantom type even though you mentioned it ^^.

If we had:
```scala
@PLenses case class Foo[A, B](x: (A, B), y: Int)
```

will it generate:
```scala
object Foo {
  def x[A, B, A$, B$]: PLens[Foo[A, B], Foo[A$, B$], (A, B), (A$, B$)] = ...
  def y[A, B]: PLens[Foo[A, B], Foo[A, B], Int, Int] = ... // so in that case it is a Lens[Foo[A,B], Int]
}
```

exlevan:
Yes, exactly.

julien-truffaut:
Great! If you get the time, it would be nice to implement this in a separate `@PLenses` macro.

julien-truffaut:
Thanks @exlevan

PR #284, opened 2015-11-06
update macro-compat version to 1.0.6

aoiroaoino:
ref.
https://gitter.im/milessabin/macro-compat?at=563b65c65680530d27049fa2


julien-truffaut:
:+1:

PR #285, opened 2015-11-07
Use Vector.updated for efficient, persistent indexed vector updates

spebbe:
Current implementation of indexed updates for Vector seems to be doing a full traversal and copy, which is O(n) or worse and also bypasses the nice persistent sharing properties of Vector. This small patch uses Vector.updated, which enables structural sharing and much lower time complexity.


NightRa:
LGTM :+1: 
Should we use bounds checking via `isDefinedAt` instead of catching exceptions?

spebbe:
Great! Yeah, isDefinedAt is probably less icky. It isn't used anywhere else in the project though, whereas \/.fromTryCatchThrowable is.

julien-truffaut:
Nice change. I would also be in favor of using `isDefinedAt`

PR #286, opened 2015-11-07
vectorIndex: use isDefinedAt instead of catching an exception

NightRa:
vectorIndex: use isDefinedAt instead of catching an exception.
As agreed apon in #285.


julien-truffaut:
cool :+1:

PR #289, opened 2015-11-12
Plated

### LINKED ISSUE #288: Plated optic
I'd like to see the [Plated type-class](https://hackage.haskell.org/package/lens-4.13/docs/Control-Lens-Plated.html#g:1) from Haskell lens implemented in Monocle:

```scala
trait Plated[A] {
  def plate: Traversal[A, A]
}

object Plated {
  def plate[A: Plated]: Traversal[A, A] = implicitly[Plated[A]].plate
  def children[A: Plated](a: A): List[A] = plate[A].getAll(a)
  def universe[A: Plated](a: A): Stream[A] = {
    val fold = plate[A].asFold
    def go(b: A): Stream[A] = fold.foldMap[Stream[A]](c => c #:: go(c))(b)
    go(a)
  }
  def transform[A: Plated](f: A => A)(a: A): A = plate[A].modify(f)(a)
}
```

The above might be a good start but I'd love to see functions like `rewrite`.

Some questions have to be answered, though, for example: Stream makes sense for `universe` but it might also make sense for `children` - what should we do?

puffnfresh:
To fix #288

### DIFF
--- a/core/src/main/scala/monocle/Plated.scala
+++ b/core/src/main/scala/monocle/Plated.scala
@@ -1,5 +1,6 @@
 package monocle
 
+import scala.annotation.implicitNotFound
 import scalaz._
 import scalaz.std.stream._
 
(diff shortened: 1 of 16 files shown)


core/src/main/scala/monocle/Plated.scala
```diff
@@ -0,0 +1,84 @@
...
+          as => Applicative[F].map(Traverse[S].traverse(as)(f)) {
+            // Free.roll does not exist in 7.1
+            Free.liftF(_).flatMap(identity)
+          },
+          x => Applicative[F].point(Free.point(x))
+        )
+    }
+  }
+
+  implicit def listPlated[A]: Plated[List[A]] = new Plated[List[A]] {
```
julien-truffaut: would you mind to move the instances of `Plated` to `monocle.std.x`, in that case `monocle.std.list`

puffnfresh:
Should `monocle.Plated` become `monocle.function.Plated` like `Each` is?

julien-truffaut:
yes please, that's where we put all the typeclasses even though it is a poor choice of package name

puffnfresh:
I've moved Plated to the right place now.

julien-truffaut:
great! If you have time, would you mind to add a test or an example for `rewriteOf `, `transform`, `transformOf`

puffnfresh:
@julien-truffaut I've added some specs for Plated, they're not brilliant but I think they're useful.

julien-truffaut:
They are definitely useful. Thanks!

:+1:

Issue #177, opened 2015-02-01
add optics for scalaz.Cofree


julien-truffaut:
```scala
  def head[S[_], A]: Lens[Cofree[S, A], A]
  def tail[S[_], A]: Lens[Cofree[S, A], S[Cofree[S, A]]]

  def coffreeToNel[A]: Iso[Cofree[List, A], NonEmptyList[A]]
  def coffreeToTree[A]: Iso[Cofree[Stream, A], Tree[A]]
```

bonus points: make the optics polymorphic (i.e. `PLens` and `PIso` instead of `Lens` and `Iso`)

LiamGoodacre:
Hello, I'm just about to play around with this as an entry point to the project.

When you say about using polymorphic optics, I can't see how `head` and `tail` can be generalized to `PLens` (other than trivially) given that they are only focusing on the top level of a `Cofree` structure and not every `head` and `tail` at all levels down.  (Although those could be nice `PTraversal`s to have.)

Also I don't believe `cofreeToNel` could be a valid `Iso` - The `Cofree List` has unrecoverable structure that will be forgotten.  If we pick a configuration of `Cofree List` to always convert to then we'd be vialating the `Iso` "modify identity" law: `iso modify id = id`.

A `PIso` for `cofreeToTree` will work though :smiley:

How does this sound?

~~~scala
def head[S[_], A]: Lens[Cofree[S, A], A]
def tail[S[_], A]: Lens[Cofree[S, A], S[Cofree[S, A]]]

def heads[S[_], A, B]: PTraversal[Cofree[S, A], Cofree[S, B], A, B]
def tails[S[_], A, B]: PTraversal[Cofree[S, A], Cofree[S, B], S[Cofree[S, A]], S[Cofree[S, B]]]

def cofreeToTree[A, B]: PIso[Cofree[Stream, A], Cofree[Stream, B], Tree[A], Tree[B]]
~~~

julien-truffaut:
you're completely right, `head` and `tail` cannot be a `PLens`. In fact, rather than defining `head` and `tail` manually we can create an instance of `Cons1` for `Cofree` (see `monocle.std.nel` for an example).

You're also right about `coffreeToNel `, I believe it should be using `Option` and not `List`:
```scala
def coffreeToNel[A]: Iso[Cofree[Option, A], NonEmptyList[A]]
```

[...]

PR #290, opened 2015-11-12
WIP - Cofree Optics

LiamGoodacre:
For https://github.com/julien-truffaut/Monocle/issues/177

* [x] ~~`(p)Heads` - `(P)Traversal` over the head values of a cofree traversable structure~~
* [x] `Each` instance for `Cofree`
* [x] `(p)CofreeToStream` - `(P)Iso` between `Cofree Option` and `OneAnd Stream`
* [x] `(p)CofreeToTree` - `(P)Iso` between `Cofree Stream` and `Tree`
* [x] `Cons1` instance for `Cofree`
* [x] Tests - need `Arbitrary` instance for `Cofree`

### DIFF
--- a/core/src/main/scala/monocle/std/Cofree.scala
+++ b/core/src/main/scala/monocle/std/Cofree.scala
@@ -10,8 +10,8 @@ object cofree extends CofreeOptics
 
 trait CofreeOptics {
 
-  private def toStream[A](c: Cofree[Option, A]): Stream[A] =
-    c.head #:: c.tail.fold(Stream.empty[A])(toStream[A])
+  private def toStream[A](optC: Option[Cofree[Option, A]]): Stream[A] =
+    optC.fold(Stream.empty[A])(c => c.head #:: toStream(c.tail))
 
   private def fromStream[A, B](z: A, c: Stream[A]): Cofree[Option, A] = c match {
     case head #:: tail => Cofree.delay(z, Some(fromStream(head, tail)))
@@ -22,7 +22,7 @@ trait CofreeOptics {
   def pCofreeToStream[A, B]: PIso[Cofree[Option, A], Cofree[Option, B],
                                   OneAnd[Stream, A], OneAnd[Stream, B]] =
     PIso[Cofree[Option, A], Cofree[Option, B], OneAnd[Stream, A], OneAnd[Stream, B]](
-      (c: Cofree[Option, A]) => OneAnd[Stream, A](c.head, toStream(c))
+      (c: Cofree[Option, A]) => OneAnd[Stream, A](c.head, toStream(c.tail))
     ){ case OneAnd(head, tail) => fromStream(head, tail) }
 
   /** [[Iso]] variant of [[pCofreeToStream]]  */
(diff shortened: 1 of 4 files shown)


puffnfresh:
Awesome stuff! :+1:

core/src/main/scala/monocle/std/Cofree.scala
```diff
@@ -0,0 +1,74 @@
...
+import scalaz.Cofree._
+import scalaz.{Cofree, Free, Applicative, Traverse, OneAnd, Tree}
+
+object cofree extends CofreeOptics
+
+trait CofreeOptics {
+
+  /** Polymorphic traversal over all the head values of a cofree structure
+    * assuming the cofree's branching functor is traversable */
+  def pHeads[S[_] : Traverse, A, B]: PTraversal[Cofree[S, A], Cofree[S, B], A, B] =
```
julien-truffaut: I believe `pHeads` go through all elements of `Cofree[S, A]`. If that's correct, we could instead create an instance of `Each` for `Cofree`. It is only monomorphic but it is relatively easy to create a `Traversal` or `PTraversal` from `Traverse` using `fromTraverse` https://github.com/julien-truffaut/Monocle/blob/master/core/src/main/scala/monocle/Traversal.scala#L189

core/src/main/scala/monocle/std/Cofree.scala
```diff
@@ -0,0 +1,74 @@
...
+  private def toStream[A](c: Cofree[Option, A]): Stream[A] =
+    c.head #:: c.tail.fold(Stream.empty[A])(toStream[A])
+
+  private def fromStream[A, B](z: A, c: Stream[A]): Cofree[Option, A] = c match {
+    case head #:: tail => Cofree.delay(z, Some(fromStream(head, tail)))
+    case _ => Cofree(z, None: Option[Cofree[Option, A]])
+  }
+
+  /** Polymorphic isomorphism between `Cofree[Option, _]` and `OneAnd[Stream, _]` */
+  def pCofreeToStream[A, B]: PIso[Cofree[Option, A], Cofree[Option, B],
```
julien-truffaut: nice, you're right it is a non empty stream

LiamGoodacre:
Thanks for reviewing.  I've added the `Each` instance to replace `pHeads`/`heads`.

I'm having trouble with the tests - I have imported `scalaz.Cofree._` which I thought would give me its `Equal` instance.  However it doesn't seem to be able to find it.
They also require an `Arbitrary` instance, should I just define this in the test file? or is there an existing instance I can use?

julien-truffaut:
I didn't realise that `Equal` and `Arbitrary` instance for `Cofree` were a bit tricky. I managed to add them following what people did in scalaz (https://github.com/scalaz/scalaz/blob/fd8176ca88c868f11b6d97149f51ec5d8b544aac/tests/src/test/scala/scalaz/CofreeTest.scala). 

Do you mind to give me access to your fork? I can submit a PR there.

xuwei-k:
> Equal and Arbitrary instance for Cofree were a bit tricky

I was wrong :( https://github.com/scalaz/scalaz/pull/1051

FYI
https://github.com/typelevel/shapeless-contrib/blob/v0.4/scalaz/main/scala/cofree.scala

julien-truffaut:
cool we can simplify the definition when scalaz 7.2 is released. In the meantime, we can always pass the natural transformation explicitly

LiamGoodacre:
@julien-truffaut awesome, I've added you as a collaborator.

julien-truffaut:
:+1:

LiamGoodacre:
:smile: Is everything fine with respect to code style, commenting, naming, etc?

julien-truffaut:
yes everything is good to me. I generally don't use extract method like `toStream` and `fromStream` but there is nothing against it.

I will wait a day to see if anyone has something to say and then I am planning to merge it.

Thank you very much for your contribution!

LiamGoodacre:
Thank you :grinning:

PR #291, opened 2015-11-15
WIP At rework

julien-truffaut:
- Updated `At` to have `A` as a target instead of `Option[A]`
   - Created a `monocle-refine` module with a dependency to `refined`
   - Created `At` instance for `Byte`, `Char`, `Int` and `Long` using @fthomas gist: https://github.com/fthomas/refined/blob/a5be764ee04200cb8e1a01f7724a022f86d0559c/core/jvm/src/main/scala/eu/timepit/refined/SafeAt.scala
   - Removed `Index` instances for `Byte`, `Char`, `Int` and `Long`

Outstanding issue:
   - `At` instances for primitive do not play well with `apply` syntax see `AtExample`


example/src/test/scala/monocle/function/AtExample.scala
```diff
@@ -34,4 +38,21 @@ class AtExample extends MonocleSuite {
     (Set(1, 2, 3) applyLens at(2) set None)     shouldEqual Set(1, 3)
   }
 
+  test("at creates a Lens from Int to one of its bit") {
+    (3 applyLens at(0: IntBits) get) shouldEqual true   // true  means bit is 1
```
fthomas: Is the `IntBits` ascription necessary? The `auto._` import should take care of converting `Int` to `IntBits`, right?

example/src/test/scala/monocle/function/AtExample.scala
```diff
@@ -34,4 +38,21 @@ class AtExample extends MonocleSuite {
...
+    (4 applyLens at(0: IntBits) get) shouldEqual false  // false means bit is 0
+
+
+    (32 applyLens at(0: IntBits) set true)   shouldEqual 33
+    (3  applyLens at(1: IntBits) modify(!_)) shouldEqual 1 // toggle 2nd bit
+
+    illTyped("""0 applyLens at(79) get""")
```
fthomas: One underused feature of `illTyped` (even in shapeless) is that it takes as optional second argument a regular expression that is matched against the compiler error. This guards against syntax errors in the code which also let `illTyped` succeed. [Here](https://github.com/fthomas/refined/blob/28d1e3802b0a1d00e3b848668e35993e88d4fc5f/core/shared/src/test/scala/eu/timepit/refined/RefineMSpec.scala#L71) is an example.

julien-truffaut: nice, I will give it a try.

example/src/test/scala/monocle/function/AtExample.scala
fthomas: Just playing with that. It seems the ascription is necessary because it guides the implicit search for the correct `At` instance.

julien-truffaut: yes I think the ascription is unfortunately needed because the type inference infer needs to know both the type of the source and the type of the index. We know the source by using `applyLens` but the index cannot be infer to the right type and so the `auto._` is not used :(

fthomas:
This has not been changed in this PR, but I just noticed that `Bits[Long].bitSize` is 32 instead of 64.

julien-truffaut:
@fthomas good catch

fthomas:
The bits (!) and pieces I understand look good to me. :+1:

I would probably have called the subproject `monocle-refined` (like the circe subproject that adds support for refined types is called `circe-refined`) but that is not my call.

julien-truffaut:
I am fine to rename the module to `monocle-refined`. However, it might cause some import confusion if we rename the package from `monocle.refine` to `monocle.refined`. WDYT?

fthomas:
I've seen problems with package names that are a root package in another project. For example if `refined` would contain a `monocle` package and I want to import `monocle` simultaneously from both projects, I'd have to prefix the monocle import with `_root_`. I haven't seen any problems with packages that have different prefixes (as `monocle.refined` and `eu.timepit.refined` would have).

julien-truffaut:
I thought it would work on the same way root package or not but scala never stop to surprise me! I can rename module and package, once it is merged, we can check there is no package conflict between refined and monocle snapshot.

example/src/test/scala/monocle/function/AtExample.scala
fthomas: Ha! In the passing examples above the ascription `: IntBits` is necessary to guide the implicit search. So this example probably fails because it can't find a suitable `At` instance and not because `79` is not a valid `IntBits`. 

Matching against the expected compile error would have caught that. :smiley:

core/src/main/scala/monocle/std/Set.scala
```diff
@@ -11,7 +11,7 @@ trait SetOptics {
     def empty = Prism[Set[A], Unit](s => if(s.isEmpty) Some(()) else None)(_ => Set.empty[A])
   }
 
-  implicit def atSet[A]: At[Set[A], A, Unit] = new At[Set[A], A, Unit] {
+  implicit def atSet[A]: At[Set[A], A, Option[Unit]] = new At[Set[A], A, Option[Unit]] {
```
NightRa: Just as a note; Option[Unit] = Boolean. Could be a possible change, but as a seperate PR as this breaks compatability.

NightRa: We break compat anyway so it doesn't matter :)

NightRa:
Most uses of `At` used `Option[A]` and not `A`.
Maybe we should seperate them?
Other than that, LGTM :+1:

core/src/main/scala/monocle/std/Set.scala
julien-truffaut: very good point. In fact, I would say it is an argument for not separating `At`.

PR #293, opened 2015-11-17
Correct Bits[Long].bitSize

fthomas:
As noted in https://github.com/julien-truffaut/Monocle/pull/291#issuecomment-157000031


julien-truffaut:
Thanks

PR #294, opened 2015-11-17
Rename subproject and package to refined

fthomas:
It seems that this doesn't cause any problems in `AtExample.scala` where we have both imports:
```scala
import eu.timepit.refined._
import monocle.refined._
```


julien-truffaut:
:+1: great

PR #300, opened 2015-11-26
Generate fresh type-parameter name for modifyF

### LINKED ISSUE #299: Bad macro hygiene
Just got this from a coworker:

```
[error] Services.scala:11: type parameter F defined in method modifyF shadows type F defined in method errHandler. You may want to rename your type parameter, or possibly remove it.
[error] @Lenses case class Services[F[_]](
[error]  ^
[error] one error found
[error] (services/compile:compileIncremental) Compilation failed
```

We have a lot of scalac warnings enabled, and we force them to be errors. Changing `F[_]` to `G[_]` makes it work.

Macros should generate a fresh identifier for the type-parameter.

puffnfresh:
Fixes #299

### DIFF
--- a/macro/src/main/scala/monocle/macros/internal/Macro.scala
+++ b/macro/src/main/scala/monocle/macros/internal/Macro.scala
@@ -72,7 +72,7 @@ private[macros] class MacroImpl(val c: blackbox.Context) {
       .find(_.name.decodedName.toString == strFieldName)
       .getOrElse(c.abort(c.enclosingPosition, s"Cannot find constructor field named $fieldName in $sTpe"))
 
-    val F = c.universe.TypeName(c.freshName("F"))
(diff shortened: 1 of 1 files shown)


jedws:
:+1:

NightRa:
It's failing to build on 2.10

julien-truffaut:
I think you can use :
```scala
TypeName(c.freshName)
```

https://github.com/milessabin/macro-compat/blob/master/core/src/main/scala_2.10/macrocompat/macrocompat.scala#L157

puffnfresh:
Think I fixed 2.10 but broke 2.11 :horse:

puffnfresh:
@julien-truffaut trying that, thanks!

puffnfresh:
There we go :fireworks:

julien-truffaut:
:+1: Thanks, I didn't know about macro hygiene.

PR #303, opened 2015-11-28
upgrade scalaz to 7.2.0-RC1 as well as test dependencies

julien-truffaut:


core/src/main/scala/monocle/std/NonEmptyList.scala
```diff
@@ -4,23 +4,23 @@ import monocle.function._
...
 
 object nel extends NonEmptyListOptics
 
 trait NonEmptyListOptics {
 
   final def pNelToOneAnd[A, B]: PIso[NonEmptyList[A], NonEmptyList[B], OneAnd[List,A], OneAnd[List,B]] =
-    PIso((nel: NonEmptyList[A])    => OneAnd[List,A](nel.head, nel.tail))(
-      (oneAnd: OneAnd[List, B]) => NonEmptyList.nel(oneAnd.head, oneAnd.tail))
+    PIso((nel: NonEmptyList[A])    => OneAnd[List,A](nel.head, nel.tail.toList))(
+      (oneAnd: OneAnd[List, B]) => NonEmptyList.nel(oneAnd.head, IList.fromList(oneAnd.tail)))
```
NightRa: Maybe make an `Iso[NonEmptyList,IList]` and then compose with `Iso[IList,List]`?

julien-truffaut: good point

julien-truffaut: in fact, here it is a bit annoying because we need to lift a `Iso[IList[A], List[A]]` to an `Iso[OneAnd[IList,A], OneAnd[List,A]]`. It is possible, we can compose with `second` and `GenIso.fields` but we don't have dependency on macro.

I updated other `Iso` with your suggestions

NightRa: :+1:

NightRa:
LGTM :+1:

PR #313, opened 2015-12-13
Add first, reverse and iso for Tuple1

exlevan:


julien-truffaut:
nice, I didn't know they were a `Tuple1`. 

:+1: once travis is done

PR #316, opened 2015-12-14
Faq

julien-truffaut:


docs/src/main/tut/faq.md
```diff
@@ -0,0 +1,108 @@
...
+
+Since `index` is weaker than `at`, we can implement an instance of `Index` on more data structure than `At`. 
+For instance, `List` or `Vector` only have an instance of `Index` because there is no way to insert an element at an 
+arbitrary index of a sequence.
+
+Note: `root` is a trick to help type inference. Without it, we would get the following error
+
+```tut:fail
+index("two").set(0)(m) 
+```
```
NightRa: Will `tut` give the error here?

NightRa:
Looks great for a start on the FAQ! :+1:

docs/src/main/tut/faq.md
julien-truffaut: yes it does. There is a limitation on the `fail` modifier where you need to use only failing code that's why I had to split examples into two `tut`, one for import and one for failing code

julien-truffaut:
Thanks, I will merge and update the website. You will see the result

PR #319, opened 2015-12-19
avoid publish when pull request

xuwei-k:


julien-truffaut:
:+1:  Thanks @xuwei-k

PR #320, opened 2015-12-19
fix implicitNotFound message of Plated

aoiroaoino:
before:
```scala
scala> monocle.function.Plated.children(100)
<console>:11: error: Could not find an instance of Plated[{A}], please check Monocle instance location policy to find out which import is necessary
       monocle.function.Plated.children(100)
                                       ^
```

after:
```scala
scala> monocle.function.Plated.children(100)
<console>:11: error: Could not find an instance of Plated[Int], please check Monocle instance location policy to find out which import is necessary
       monocle.function.Plated.children(100)
                                       ^
```


NightRa:
:+1:

puffnfresh:
Thanks! :+1:

Issue #330, opened 2016-03-10
2.11.8 support


acruise:
It'd be nice to have a new build of 1.2.0 with macro support for 2.11.8. Thanks!

julien-truffaut:
what would be the difference if the lib is compiled against 2.11.7 or 2.11.8?

aoiroaoino:
related to macro's issue is following...?
https://issues.scala-lang.org/browse/SI-9616?jql=project%20%3D%20SI%20AND%20resolution%20%3D%20Fixed%20AND%20fixVersion%20%3D%20%22Scala%202.11.8%22%20AND%20text%20~%20%22macro%22%20ORDER%20BY%20component%20ASC%2C%20priority%20DESC
hmm..... I think that necessarily there is no need to release.

However, https://github.com/sbt/sbt/issues/2076 has been fixed in v2.11.8.
We can try to use `-Ybackend:GenBCode` option. :smile:

julien-truffaut:
@aoiroaoino do you all the scalac options we need to benefit from java8? It would be nice to give it a try.

xuwei-k:
> all the scalac options we need to benefit from java8

- https://github.com/scalaz/scalaz/commit/e3f01cdfb399412b400b9b2250ad755516d7ab2f
- https://github.com/scala/make-release-notes/blob/9cfbdc8c92f94/experimental-backend.md#emitting-java-8-style-lambdas
- https://github.com/lrytz/experimental-backend-2.11/blob/76b7a45b73a08e5b17641d3fc179e9861bba2f80/build.sbt

julien-truffaut:
Awesome, thanks @xuwei-k

julien-truffaut:
can I close it? Or someone need a release?

aoiroaoino:
I think that this issue has finished :+1:

PR #333, opened 2016-03-12
add scalac option for java8

julien-truffaut:
see #330

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ scala:
 - 2.11.8
 - 2.10.6
 jdk:
-- oraclejdk7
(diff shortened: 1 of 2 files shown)


xuwei-k:
Also, `scala-java8-compat` dependency need to use invokedynamic and LambdaMetaFactory.

```scala
libraryDependencies += "org.scala-lang.modules" %% "scala-java8-compat" % "0.7.0"
```

julien-truffaut:
cheers I missed the dependency

aoiroaoino:
awesome :+1: :+1: :+1: 

thanks @xuwei-k :wink:

PR #335, opened 2016-03-31
Scala.JS support

japgolly:


japgolly:
`BigDecimalSpec` found a bug in Scala.JS (https://github.com/scala-js/scala-js/issues/2314) and so has been temporarily excluded from JS tests.

Apart from that, it's all good. Everything else passes. Travis is happy.

ngbinh:
:+1: awesome!

xuwei-k:
Should we use `CrossType.Pure` or custom CrossType for avoid move files?
- https://github.com/scala-js/scala-js/blob/v0.6.8/sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/cross/CrossProject.scala#L83-L98
- https://github.com/scala-js/scala-js/blob/v0.6.8/sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/cross/CrossType.scala#L49-L55
- https://github.com/scalaz/scalaz/blob/v7.2.1/project/build.scala#L85-L96
- https://github.com/milessabin/shapeless/blob/shapeless-2.3.0/build.sbt#L85-L91

japgolly:
I think there's no benefit in avoiding the move of files. If Monocle ever
needs to provide or include/exclude functionality by platform, then it's
ready to go (like how I temporarily disabled `BigDoubleSpec`). Git handles
renames well, we're using the std structure that people expect for SJS
projects, I see all up-sides here and no down-sides.

julien-truffaut:
Awesome, thanks @japgolly

julien-truffaut:
btw I think it is better to move the files, it will be more consistent with other project using scala.js

ngbinh:
:clap:

japgolly:
You're welcome! Thank you for Monocle. :D

ngbinh:
@julien-truffaut can we have a new release with scala.js cross build? Thanks

julien-truffaut:
yes we can, I am just wandering about support for java 8 from https://github.com/julien-truffaut/Monocle/pull/333 

Shall we wait for `1.3.0` to support java 8 and release scala.js support in `1.2.1`? WDYT

ngbinh:
I wouldn't mind having scala.js support version of monocle now. We are already on `java8` but if it will be a short while then we can wait. :)

julien-truffaut:
@japgolly I am seeing the following error when I try to build the project locally, any idea?

```
[error] a module is not authorized to depend on itself: com.github.julien-truffaut#monocle_2.11;1.3.0-SNAPSHOT
java.lang.IllegalArgumentException: a module is not authorized to depend on itself: com.github.julien-truffaut#monocle_2.11;1.3.0-SNAPSHOT
  at org.apache.ivy.core.resolve.IvyNodeCallers.addCaller(IvyNodeCallers.java:167)
  at org.apache.ivy.core.resolve.IvyNode.addCaller(IvyNode.java:1076)
```

japgolly:
Hi. I just woke up. Give me a few min and I'll have a look
otherwise-immediately :)

japgolly:
Thanks for waiting. I had basically a typo. Creating new PR now....
*yaaaaaawn* :sleepy:

PR #340, opened 2016-04-09
336 - stringToBoolean to respect Prism laws

hasumedic:
This PR makes the String to Boolean Prism to respect Prism laws by constraining the string input to only be lowercase.

This also removes the usage of Scalaz's Disjunction since Exceptions are not being raised any more.


NightRa:
Looks good, thanks :+1:

hasumedic:
@pvillega was too quick! Now it looks better

NightRa:
Thanks!

PR #341, opened 2016-04-09
Add sans to At

dabd:
Add function 'sans' to delete a value associated with a key in a Map-like container.


julien-truffaut:
Thanks :+1:

NightRa:
I think we should give it a more intuitive name such as `remove` or `delete`

julien-truffaut:
make sense, I think `sans` came from haskell lens but we can change it.

PR #346, opened 2016-04-10
[DONT MERGE] Release 1.2.1

julien-truffaut:
tentative release for 1.2.1

Potential issue, we made a major upgrade of shapeless (2.2.5 -> 2.3.0). Shall we revert shapeless to 2.2.5?

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ resolvers += Resolver.sonatypeRepo("releases")
 resolvers += Resolver.sonatypeRepo("snapshots")
 
 val scalaVersion   = "2.11.8"    // or "2.10.6"
-val libraryVersion = "1.2.0"     // or "1.3.0-SNAPSHOT"
+val libraryVersion = "1.2.1"     // or "1.3.0-SNAPSHOT"
 
 libraryDependencies ++= Seq(
   "com.github.julien-truffaut"  %%  "monocle-core"    % libraryVersion,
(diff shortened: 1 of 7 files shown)


julien-truffaut:
we need to revert shapeless to 2.2.5 and refined to 0.3.2

julien-truffaut:
@aoiroaoino @NightRa @japgolly If you have a min, would you mind to review this release? I run mima and everything seems to be backward compatible with 1.2.0.

japgolly:
Most libraries are backwards-compatible in their minor releases but rarely forwards-compatible. I tend to keep the *z* in versions *x.y.z* low in library deps. Example I just released scalajs-react the other day and I used  7.2.1 for official Scala.JS support but ignored the 7.2.2 release because I want users to be able to use 7.2.1+ not 7.2.2+. It might be a good policy to go with the minimum major versions as a library. Up to you.

Next, I've not used mima so I don't know about those settings.

Otherwise LGTM :+1: .
Just remember to change the date in the changelog before you hit the release button. :smile:

julien-truffaut:
thanks @japgolly I didn't think about using low patch version in a library for the arguments you mentioned. However, I think it is more convenient to get high patch version such as people get the latest bug fixes transitively. I am more than open to change that if it makes sense for users.

Good point about the release date!

julien-truffaut:
Release is done, I will cherry-pick the commit back to master. @NightRa suggested to use `git rebase -i` next time.

NightRa:
:+1:

PR #347, opened 2016-04-10
Introduce generic Each derivation

Astrac:
Implement a generic `Each` instance for any product tuple and case class which wraps all entities of the same type using shapeless typeclass derivation.

I didn't put the tests in `TupleNSpec` because I realised after leaving the hack-day that this would have worked with any product type, including case classes.

### DIFF
--- a/generic/shared/src/main/scala/monocle/generic/Product.scala
+++ b/generic/shared/src/main/scala/monocle/generic/Product.scala
@@ -1,11 +1,35 @@
 package monocle.generic
 
-import monocle.Iso
+import language.higherKinds
+import monocle.PTraversal
+import monocle.function.Each
+import monocle.{ Iso, Traversal }
 import monocle.generic.internal.TupleGeneric
+import scalaz.Applicative
+import scalaz.syntax.apply._
+import shapeless.{ ::, Generic, HList, HNil }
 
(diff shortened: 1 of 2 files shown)


generic/shared/src/main/scala/monocle/generic/Product.scala
```diff
@@ -1,11 +1,37 @@
...
 object product extends ProductOptics
 
 trait ProductOptics {
   def productToTuple[S <: Product](implicit ev: TupleGeneric[S]): Iso[S, ev.Repr] =
     Iso[S, ev.Repr](s => ev.to(s))(t => ev.from(t))
-}
\ No newline at end of file
+
+  implicit def hNilEach[A] = new Each[HNil, A] {
+    def each: Traversal[HNil, A] = new PTraversal[HNil, HNil, A, A] {
```
julien-truffaut: I think we could replace this by `Traversal.void[HNil, A]` as it is a noop traversal

julien-truffaut:
it looks really good :)

Astrac:
Thanks for the suggestion, I updated my commit to use `Traversal.void[HNil, A]`.

julien-truffaut:
Great, thanks @Astrac :+1:

PR #352, opened 2016-04-22
add release notes about non backward compatible change of mega imports

julien-truffaut:


NightRa:
LGTM :+1:

NightRa:
Have we stopped putting the release notes in `note/`?

julien-truffaut:
tbh I had forgotten about  `note/`, maybe we should move the old release notes to the web site or simply delete them. I don't think anyone care about `0.x` release.

PR #354, opened 2016-04-30
Treat PPrisms as smart constructors for type T.

sellout:
We use this at @slamdata to avoid having to deal with various inference problems caused by subtyping. @wemrysi came up with this approach.

### DIFF
--- a/test/shared/src/test/scala/monocle/Arities.scala
+++ b/test/shared/src/test/scala/monocle/Arities.scala
@@ -4,5 +4,5 @@ sealed trait Arities
 final case class Nullary() extends Arities
 final case class Unary(i: Int) extends Arities
 final case class Binary(s: String, i: Int) extends Arities
-final case class Quintary(u: Unit, b: Boolean, s: String, i: Int, f: Double)
+final case class Quintary(a: Any, b: Boolean, s: String, i: Int, f: Double)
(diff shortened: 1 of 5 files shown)


julien-truffaut:
Great! Could you add a test for a few `apply` variants (e.g. 0, 1 and x arguments)?
Would you mind to add the same `apply` method for `Iso` as well.

sellout:
Sorry this took so long. Happy to make more changes, squash/rebase/whatever.

sellout:
Apparently 2.10 doesn’t like `Unit`. I’m on it.

test/shared/src/test/scala/monocle/Arities.scala
```diff
@@ -4,5 +4,5 @@ sealed trait Arities
 final case class Nullary() extends Arities
 final case class Unary(i: Int) extends Arities
 final case class Binary(s: String, i: Int) extends Arities
-final case class Quintary(u: Unit, b: Boolean, s: String, i: Int, f: Double)
```
julien-truffaut: can you use another type than `Any`? e.g. `Char`

julien-truffaut:
@sellout looks great

julien-truffaut:
👍 Thanks you @sellout

PR #357, opened 2016-05-24
Revisit laws and fix Optional.first and second

julien-truffaut:


build.sbt
```diff
@@ -56,6 +56,8 @@ lazy val macroCompat = Def.setting("org.typelevel" %%% "macro-compat" % "1.1.0")
 lazy val macroVersion = "2.1.0"
 lazy val paradisePlugin = compilerPlugin("org.scalamacros" %  "paradise"       % macroVersion cross CrossVersion.full)
 
+lazy val kindProjector = "org.spire-math" % "kind-projector" % "0.7.1" cross CrossVersion.binary
```
aoiroaoino: coooool 👍

julien-truffaut:
@aoiroaoino @NightRa when you get the time, would you mind to have a look at this PR.

I am working on upgrading scalacheck such as we can benefit from the new `CoGen` and so have good random function. However, I would prefer to do that in a separate PR.

core/shared/src/main/scala/monocle/law/LensLaws.scala
```diff
@@ -14,12 +15,18 @@ class LensLaws[S, A](lens: Lens[S, A]) {
   def setGet(s: S, a: A): IsEq[A] =
     lens.get(lens.set(a)(s)) <==> a
 
-  def setIdempotent(s: S, a: A): IsEq[S] =
-    lens.set(a)(lens.set(a)(s)) <==> lens.set(a)(s)
```
aoiroaoino: Why remove it?

aoiroaoino:
Sorry for the late reply. 😅 
I have an above question. other than that looks good. 👍

core/shared/src/main/scala/monocle/law/LensLaws.scala
julien-truffaut: I believe it is redundant with `modifyId` and `consistentSetModify` but I might be wrong

julien-truffaut: In fact, I think I am wrong. I don't think `setIdempotent` is redundant. I will add it back

aoiroaoino:
👍 👍 👍

PR #358, opened 2016-05-26
Fixed typo in optics.md

amrhassan:
Fixed typo


julien-truffaut:
Good catch, thank you 👍

Issue #179, opened 2015-02-04
Attempt to use Prism as more generic pattern match


julien-truffaut:
cf haskell total
http://www.haskellforall.com/2015/01/total-100-exhaustive-pattern-matching.html

puffnfresh:
I've got something working but need to come up with a version which gives some type-inference. @julien-truffaut should this go into a Monocle module or just be a separate library?

julien-truffaut:
Wow that's awesome! It could go either way, I might be slightly in favour of a module as it makes versioning easier to manage but if you prefer to release it in a separate project, it is good for me as well.

NightRa:
I'd prefer it to be a part of Monocle. It can go as another module if desired.
Kudos @puffnfresh!

aloiscochard:
I'll looking forward to see your imp. @puffnfresh! well done :)

julien-truffaut:
@puffnfresh by any chance, do you have your current work pushed somewhere? I am really curious to see how it looks like.

sellout:
Why was this closed? I feel like #361 doesn’t quite fulfill the promise of Haskell’s total.

julien-truffaut:
You're right, I have been too eager to close it.

PR #361, opened 2016-06-06
Add unapply method to PPrism and PIso.

### LINKED ISSUE #360: Add `unapply` to Prism and Iso.
I got `apply` added to these classes, and now I want the reverse!

I commonly write pattern-matching helpers like
```scala
  object BsonType {
    def apply(name: String): Bson = Bson.Doc("$mongodb.type" -> name)
    def unapply(value: Bson): Option[String] = value match {
      case Bson.Doc(map) => if (map.size == 1) map.get("$mongodb.type") else None
      case _             => None
    }
  }
```
and this looks suspiciously like a Prism. I’d like to make the relationship between apply/unapply more explicit and do
```scala
  val BsonType = new Prism[Bson, String] {
    case Bson.Doc(map) => if (map.size == 1) map.get("$mongodb.type") else None
    case _             => None
  } (Bson.Doc("$mongodb.type" -> _))
```

The same can be done with `Iso`, modulo `unapply` returning `Some` instead of `Option` to avoid breaking exhaustiveness checking.

sellout:
Fixes #360.

The goal here is to be able to use arbitrary Prisms and Isos as
apply/unapply objects, making the relationship between apply/unapply
more explicit.

### DIFF
--- a/core/shared/src/main/scala/monocle/Iso.scala
+++ b/core/shared/src/main/scala/monocle/Iso.scala
@@ -279,6 +279,8 @@ abstract class PIso[S, T, A, B] extends Serializable { self =>
 
   def apply[C, D, E, F, G, H](c: C, d: D, e: E, f: F, g: G, h: H)(implicit ev: (C, D, E, F, G, H) <~< B): T =
     apply(ev((c, d, e, f, g, h)))
+
+  def unapply(obj: S): Some[A] = Some(get(obj))
 }
 
 object PIso extends IsoInstances {
--- a/core/shared/src/main/scala/monocle/Prism.scala
+++ b/core/shared/src/main/scala/monocle/Prism.scala
@@ -253,6 +253,8 @@ abstract class PPrism[S, T, A, B] extends Serializable { self =>
 
   def apply[C, D, E, F, G, H](c: C, d: D, e: E, f: F, g: G, h: H)(implicit ev: (C, D, E, F, G, H) <~< B): T =
     apply(ev((c, d, e, f, g, h)))
+
+  def unapply(obj: S): Option[A] = getOption(obj)
 }
 
 object PPrism extends PrismInstances {
--- a/test/shared/src/test/scala/monocle/PrismSpec.scala
+++ b/test/shared/src/test/scala/monocle/PrismSpec.scala
@@ -76,10 +76,10 @@ class PrismSpec extends MonocleSuite {
   }
 
   test("unapply") {
-    (Nullary() match { case _nullary(unit) => unit }) shouldEqual (())
-    (Unary(3) match { case _unary(value) => value * 2 }) shouldEqual 6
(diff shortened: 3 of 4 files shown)


test/shared/src/test/scala/monocle/PrismSpec.scala
```diff
@@ -74,4 +74,13 @@ class PrismSpec extends MonocleSuite {
     _quintary('x', true, "bar", 13, 0.4) shouldEqual
       Quintary('x', true, "bar", 13, 0.4)
   }
+
+  test("unapply") {
+    (Nullary() match { case _nullary(unit) => unit }) shouldEqual (())
```
julien-truffaut: could you upcast `Nullary` and other constructors to `Arities`?

core/shared/src/main/scala/monocle/Prism.scala
```diff
@@ -253,6 +253,8 @@ abstract class PPrism[S, T, A, B] extends Serializable { self =>
 
   def apply[C, D, E, F, G, H](c: C, d: D, e: E, f: F, g: G, h: H)(implicit ev: (C, D, E, F, G, H) <~< B): T =
     apply(ev((c, d, e, f, g, h)))
+
+  def unapply(obj: S): Option[A] = getOption(obj)
```
julien-truffaut: Nice, I didn't know this trick

core/shared/src/main/scala/monocle/Iso.scala
```diff
@@ -279,6 +279,8 @@ abstract class PIso[S, T, A, B] extends Serializable { self =>
 
   def apply[C, D, E, F, G, H](c: C, d: D, e: E, f: F, g: G, h: H)(implicit ev: (C, D, E, F, G, H) <~< B): T =
     apply(ev((c, d, e, f, g, h)))
+
+  def unapply(obj: S): Some[A] = Some(get(obj))
```
julien-truffaut: Nice, I didn't know this trick

julien-truffaut:
@puffnfresh do you have any opinion as it relates to #179 but without totality checking

core/shared/src/main/scala/monocle/Iso.scala
puffnfresh: Yes, this is very useful for Iso!

puffnfresh:
@julien-truffaut my opinion is that this is good. I think there is exhaustivity checking on Iso due to the seemingly undocumented "unapply refined to Some" trick.

sellout:
Uhh, apparently pocket-closing is up there with pocket dialing these days?

core/shared/src/main/scala/monocle/Iso.scala
sellout: @nuttycom showed me this trick a while ago, and I’ve been loving it ever since.

julien-truffaut:
👍 Thank you @sellout, `apply` and `unapply` on `Iso` and `Prism` are a great addition!

PR #365, opened 2016-06-09
Initial Wrapped function

puffnfresh:


example/src/test/scala/monocle/function/WrappedExample.scala
```diff
@@ -0,0 +1,26 @@
...
+import scalaz.{@@, Tags}
+import scalaz.std.anyVal._
+import scalaz.std.list._
+import scalaz.std.function._
+import scalaz.syntax.foldable._
+
+class WrappedExample extends MonocleSuite {
+
+  test("wrapped is an Iso") {
+    (Tags.Max(100) applyIso wrapped[Int @@ Tags.Max, Int] get) shouldEqual 100
```
julien-truffaut: are the type annotations necessary? if yes, it really limits the usefulness of `wrapped`

example/src/test/scala/monocle/function/WrappedExample.scala
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  test("unwrapped is an Iso") {
+    ("Hello" applyIso unwrapped[String @@ Tags.Dual, String] get) shouldEqual Tags.Dual("Hello")
+  }
+
+  test("ala wraps and unwraps inside of a functor") {
+    ala[Boolean @@ Tags.Disjunction, Boolean, ({type λ[α] = List[Boolean] => α})#λ](f => xs => xs.foldMap(f)).apply(List(true, false, true)) shouldEqual true
```
julien-truffaut: can you use kind-projector here? e.g. `List[Boolean] => ?`

example/src/test/scala/monocle/function/WrappedExample.scala
puffnfresh: No, they're not - I added them in while working on it and forgot to remove them. Thanks.

example/src/test/scala/monocle/function/WrappedExample.scala
puffnfresh: kind-projector is not enabled for the examples sub-project. I'm thinking that we should probably have a specialised `ala` combinator so that we don't need explicit types at all here.

julien-truffaut: what do you mean by specialised `ala`?

core/shared/src/main/scala/monocle/function/Wrapped.scala
```diff
@@ -0,0 +1,26 @@
+package monocle.function
+
+import monocle.Iso
+
+import scalaz.Functor
+
+import scala.annotation.implicitNotFound
+
+/**
+  * Typeclass that defines an [[Iso]] from an `S` to an `A`
```
julien-truffaut: Could you add something like `S` is expected to wrap an `A`

core/shared/src/main/scala/monocle/Iso.scala
```diff
@@ -96,6 +96,9 @@ abstract class PIso[S, T, A, B] extends Serializable { self =>
   @inline final def product[S1, T1, A1, B1](other: PIso[S1, T1, A1, B1]): PIso[(S, S1), (T, T1), (A, A1), (B, B1)] =
     split(other)
 
+  @inline final def au[F[_]: Functor](f: (B => T) => F[S]): F[A] =
```
julien-truffaut: do you have an example of `au` other than a helper for `ala`?

puffnfresh:
@julien-truffaut pushed both changes.

julien-truffaut:
Awesome. Yes I also feared that `ala` and `au` would not be very helpful with scala type inference :(

Thanks! 👍

PR #387, opened 2016-08-04
Add State syntax for Optional

cb372:
Similar to what was already available for Lens.

I'm not sure how useful this is to anyone apart from me, so feel free to reject this PR.

My use case for this is concisely chaining JSON transformations using `JsonPath` in Circe.

e.g. if I have some input JSON

```scala
val inputJson = parse("""
{
  "user": {
    "name": "Chris",
    "age": 31
  },
  "timestamp": 1470326934000
}
""").getOrElse(Json.Null)
```

then I could transform it as follows:

```scala
val transform = for {
  _ <- root.user.name.string mod (_.toLowerCase)
  _ <- root.user.age.int mod (_ + 1)
  x <- root.timestamp.long mod (new DateTime(_).toString)
} yield x

val transformedJson = transform.exec(inputJson)
```

### DIFF
--- a/state/shared/src/main/scala/monocle/state/StateOptionalSyntax.scala
+++ b/state/shared/src/main/scala/monocle/state/StateOptionalSyntax.scala
@@ -19,10 +19,10 @@ final class StateOptionalOps[S, T, A, B](optional: POptional[S, T, A, B]) {
     toState
 
   /** modify the value viewed through the Optional and return its *old* value, if there was one */
-  def mod(f: A => B): IndexedState[S, T, Option[A]] =
+  def modo(f: A => B): IndexedState[S, T, Option[A]] =
     IndexedState(s => (optional.modify(f)(s), optional.getOption(s)))
 
   /** set the value viewed through the Optional and return its *old* value, if there was one */
-  def assign(b: B): IndexedState[S, T, Option[A]] = mod(_ => b)
+  def assigno(b: B): IndexedState[S, T, Option[A]] = modo(_ => b)
 
 }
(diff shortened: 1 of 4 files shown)


state/shared/src/main/scala/monocle/state/StateOptionalSyntax.scala
```diff
@@ -0,0 +1,28 @@
...
+  /** transforms a POptional into a State */
+  def toState: State[S, Option[A]] =
+    State(s => (s, optional.getOption(s)))
+
+  /** alias for toState */
+  def st: State[S, Option[A]] =
+    toState
+
+  /** modify the value viewed through the Optional and return its *old* value, if there was one */
+  def mod(f: A => B): IndexedState[S, T, Option[A]] =
```
julien-truffaut: in lens state, we called this method `modo` where the last `o` stands for original

state/shared/src/main/scala/monocle/state/StateOptionalSyntax.scala
```diff
@@ -0,0 +1,28 @@
...
+    IndexedState(s => (optional.modify(f)(s), optional.getOption(s)))
+
+  /** set the value viewed through the Optional and return its *old* value, if there was one */
+  def assign(b: B): IndexedState[S, T, Option[A]] = mod(_ => b)
```
julien-truffaut: same here

julien-truffaut:
It looks good, state support is an area which definitely needs some love. I just made a few comments regarding API consistency

julien-truffaut:
👍 Thanks Chris

PR #391, opened 2016-09-10
Added to Plated transformation counting number of changes

arkadius:
I've recently used this great library for transformation of nested structure. One thing that I founded out that was missed is the ability to count number of changes. I've prepared the function that can make it. Maybe it wil be helpfull for others. WDYT?

### DIFF
--- a/core/shared/src/main/scala/monocle/function/Plated.scala
+++ b/core/shared/src/main/scala/monocle/function/Plated.scala
@@ -66,10 +66,12 @@ trait PlatedFunctions {
     l.modify(b => transformOf(l)(f)(f(b)))(a)
 
   /** transforming counting changes */
-  def transformC[A: Plated](pf: PartialFunction[A, A])(a: A): (Int, A) = {
-    transformM[A, State[Int, ?]](
-      pf.andThen(b => State((i: Int) => (i + 1, b)))
-        .applyOrElse(_, State.state))(a).runZero
+  def transformCounting[A: Plated](f: A => Option[A])(a: A): (Int, A) = {
+    transformM[A, State[Int, ?]] { b =>
+      f(b).map(c => State((i: Int) => (i + 1, c)))
+        .getOrElse(State.state(b))
+    }(a).runZero
+
   }
 
   /** transforming every element using monadic transformation */
--- a/core/shared/src/main/scala/monocle/function/Plated.scala
+++ b/core/shared/src/main/scala/monocle/function/Plated.scala
@@ -3,7 +3,10 @@ package monocle.function
 import monocle.{Setter, Traversal}
 
 import scala.annotation.implicitNotFound
+import scalaz.{Monad, State}
 import scalaz.std.stream._
+import scalaz.syntax.monad._
+import scalaz.std.anyVal._
 
 /**
   * [[Plated]] is a type-class for types which can extract their immediate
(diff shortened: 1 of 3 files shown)


julien-truffaut:
Hi @arkadius, thanks for your contribution. Can't you achieve the same thing using `modifyF` on `plate` using the `State` applicative? Also counting the number of transformation is something that could be generalised to all `Traversal`, not only `Plated`

arkadius:
Hi. It was my first thought too, but then I found out that `modifyF` won't modify nested objects like `Plated.transform` do. The `Plated.transform` takes only function - not Monad. So I couldn't pass `State` to it. And it is why I've added `transformM` - I'm not praud of the name but don't know which will be better.
I'm not pretty sure that tranformation with counting is so common usage, maybe only transformation with `Monad` would be enough? But if so, I can add simmillar methods: `PTraversal.modifyC` and I wonder that `PSetter.modifyC`, and `PLens.modifyC` too? Don't you think that `Plens` should be a subtype of `PSetter`?

core/shared/src/main/scala/monocle/function/Plated.scala
```diff
@@ -62,4 +65,19 @@ trait PlatedFunctions {
...
-}
+  /** transforming counting changes */
+  def transformC[A: Plated](pf: PartialFunction[A, A])(a: A): (Int, A) = {
+    transformM[A, State[Int, ?]](
+      pf.andThen(b => State((i: Int) => (i + 1, b)))
+        .applyOrElse(_, State.state))(a).runZero
+  }
+
+  /** transforming every element using monadic transformation */
+  def transformM[A: Plated, M[_]: Monad](f: A => M[A])(a: A): M[A] = {
```
julien-truffaut: I think you can reduce the constraint from `Monad` to `Bind`

core/shared/src/main/scala/monocle/function/Plated.scala
```diff
@@ -62,4 +65,19 @@ trait PlatedFunctions {
   def transformOf[A](l: Setter[A, A])(f: A => A)(a: A): A =
     l.modify(b => transformOf(l)(f)(f(b)))(a)
 
-}
+  /** transforming counting changes */
+  def transformC[A: Plated](pf: PartialFunction[A, A])(a: A): (Int, A) = {
```
julien-truffaut: would you mind to replace `PartialFunction[A, A]` by `A => Option[A]` for consistency with `rewriteOf`

julien-truffaut: maybe we could make the name more explicit since this method is probably not going to be called frequently. What about `transformCounting`?

example/src/test/scala/monocle/std/TreeExample.scala
```diff
@@ -42,4 +43,10 @@ class TreeExample extends MonocleSuite {
     universe(tree) shouldEqual Stream(tree, Leaf(2), Leaf(3))
   }
 
+  test("transformC transform Tree nodes counting number of changed nodes") {
```
julien-truffaut: can you also write a property like: 
`for all xs: List[A], _1 compose transformC(Option)(xs) == (xs.size, xs)`

julien-truffaut:
I got it, I didn't notice the need of `flatMap` because `rewrite` is recursive. So it does make complete sense to define it for `Plated`.

 I just made a few comments but overhaul it looks great :)

core/shared/src/main/scala/monocle/function/Plated.scala
arkadius: It should be also an `Applicative` (for purpose of `modifyF`) so I can only reduce to `Bind with Applicative`. Do you think it will be better?

julien-truffaut: of course you're correct `Monad` is `Bind with Applicative` so all good :)

core/shared/src/main/scala/monocle/function/Plated.scala
arkadius: done

example/src/test/scala/monocle/std/TreeExample.scala
arkadius: done

julien-truffaut:
Awesome 👍

PR #396, opened 2016-09-12
Implements mapping method in Iso, to lift a Iso[A, S] to an Iso[F[A],…

mdulac:
… F[S]]


julien-truffaut:
nice :)

Even though the laws you added are completely correct, I am not sure we should include them as they will slow down the tests. Maybe we could only test `mapping` for a single optic as the implementation is fairly straightforward?

mdulac:
Sure, and that's what I did initially.
I will commit a single test with the Id Functor and remove the laws checks. Is that better ?

julien-truffaut:
perfect, thank you!

mdulac:
Done :)

julien-truffaut:
👍 thanks @mdulac

Issue #216, opened 2015-06-09
Release a version for Scala 2.12-RC2


julien-truffaut:

julien-truffaut:
blocked on https://github.com/typelevel/discipline/issues/16 which in turns is blocked on scpec2 and scalacheck upgrade

SethTisue:
ping?

julien-truffaut:
@SethTisue I was planning to wait for RC2 and shapeless release to release all monocle modules at once. If you need `monocle-core` for RC1, I can give it a try tonight.

SethTisue:
I deleted my comment because I'd forgotten about #397 :-)

to help us QA RC1, it would be good if you at least had a WIP branch that builds monocle-core, just so you can let us know about any regressions. but for actually publishing, waiting for RC2 seems fine.

julien-truffaut:
Np, I pushed `scala-2.12-RC1-core`. `core` compiles with RC1

SethTisue:
once Shapeless publishes for RC2, I hope you will too

julien-truffaut:
sure, I am waiting for it. Please don't hesitate to ping me if I miss the announcment

SethTisue:
Shapeless 2.3.2 for 2.12.0-RC2 is on Maven Central now.

julien-truffaut:
depend on https://github.com/non/kind-projector/issues/39

PR #403, opened 2016-10-20
upgrade dependencies for scala 2.12.0-RC2

julien-truffaut:


julien-truffaut:
fix #216 216

julien-truffaut:
@SethTisue 1.3.1 is on its way to maven central and it supports scala 2.12.0-RC2

PR #405, opened 2016-10-22
Refactor typeclass

julien-truffaut:
I guess this is quite a controversial change. The idea is to move all typeclass instances in `core` from `monocle.std` to the corresponding typeclass companion object. This means users will be able to use most typeclass based optics using only `import monocle.function._`.

It is actually how an early version of monocle was done, I decided to move to the `std` package approach to be more consistent between modules (e.g. `generic` or `refined` module cannot define instance in companion object of typeclass) and to be more consistent with libraries like `scalaz`. However, I believe that most usage of our typeclasses concern either vanilla scala or scalaz types, so it makes sense to simplify this use case. 

As far as I remember, @NightRa was originally against the move to std module.

### DIFF
--- a/core/shared/src/main/scala/monocle/function/At.scala
+++ b/core/shared/src/main/scala/monocle/function/At.scala
@@ -16,6 +16,13 @@ abstract class At[S, I, A] extends Serializable {
   def at(i: I): Lens[S, A]
 }
 
+trait AtFunctions {
+  def at[S, I, A](i: I)(implicit ev: At[S, I, A]): Lens[S, A] = ev.at(i)
+
+  /** delete a value associated with a key in a Map-like container */
+  def remove[S, I, A](i: I)(s: S)(implicit ev: At[S, I, Option[A]]): S =
+  ev.at(i).set(None)(s)
+}
 
 object At extends AtFunctions {
   def apply[S, I, A](get: I => S => A)(set: I => A => S => S): At[S, I, A] =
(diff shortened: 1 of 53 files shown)


aoiroaoino:
It's very useful 👍 

However, I have several concerns.
- can cause object hypertrophy by type class instances.
- For example, instances of the type class for the List would be scattered in various files.  
(but, it's a trade-off

How do you think?

julien-truffaut:
it is true that instances for a class will be scattered around several typeclasses so it will make it more difficult to know what are the typeclasses implemented for List. At the same time, it will be easier to know what are the instances for `At`, I guess as you say it is a trade-off. 

I am not sure what do you mean by object hypertrophy, is it that the companion object would be too big? If yes, I am not sure what are the consequences. Is it less readable or could it cause a slow down for implicit resolutions because more implicits would be in scope?

core/shared/src/main/scala/monocle/function/At.scala
```diff
@@ -12,10 +12,17 @@ import scala.annotation.implicitNotFound
...
+abstract class At[S, I, A] extends Serializable {
   def at(i: I): Lens[S, A]
 }
 
+trait AtFunctions {
+  def at[S, I, A](i: I)(implicit ev: At[S, I, A]): Lens[S, A] = ev.at(i)
+
+  /** delete a value associated with a key in a Map-like container */
+  def remove[S, I, A](i: I)(s: S)(implicit ev: At[S, I, Option[A]]): S =
+  ev.at(i).set(None)(s)
```
aoiroaoino: would you fix indent?

aoiroaoino:
> I guess as you say it is a trade-off.

Yes!

> Is it less readable or could it cause a slow down for implicit resolutions because more implicits would be in scope?

It is both. but, sorry, I too am not sure whether there is influence how much the future.

maybe, It is not serious problems for the time being. This change is looks good. 👍

core/shared/src/main/scala/monocle/function/At.scala
julien-truffaut: sure :)

julien-truffaut:
Since it is a major change, I propose that we do a milestone / RC release for 1.5 in order to get some feedbacks from users. We can always roll-back if we discover a major issue. Thanks for the review!

PR #406, opened 2016-10-22
Update build settings

aoiroaoino:
ref.
https://github.com/scala-js/scala-js/pull/2395
https://github.com/scala-js/scala-js/issues/2579


julien-truffaut:
👍

Issue #375, opened 2016-07-05
Monocle sticker


julien-truffaut:
If by any chance we have an artist around, it would be awesome to make a monocle sticker e.g.

https://twitter.com/typelevel/status/741299243500097536?lang=en-gb

aoiroaoino:
You mean, renew the logo?

julien-truffaut:
We can renew the logo or we could have a simplified version that fit into an hexagone.

julien-truffaut:
@DavidMorganBrow (twitter handle) proposed the following logo, I think it is really awesome:
![pasted image at 2016_11_11 11_32 am](https://cloud.githubusercontent.com/assets/1859417/20219149/ec8d121e-a81f-11e6-93cb-b3fa6e8ba822.png)

aoiroaoino:
It's so cooool! I realy like it.

Issue #382, opened 2016-07-20
Please add imports to README


nuttycom:
The lack of imports in the README means that users have to go hunting to figure out where to get the functionality. Please add them!

julien-truffaut:
Sure I will try to go through the Readme and improve the situation. In the meantime, you can have a look at the website http://julien-truffaut.github.io/Monocle/, all the examples there are compiled such as they cannot be out of date or miss an import.

PR #411, opened 2016-11-05
use sbt-microsite for docs

### LINKED ISSUE #401: replace website with sbt-microsites
see https://47deg.github.io/sbt-microsites/docs/settings.html

julien-truffaut:
fix #401 

There are still some issues:
- syntax highlighting seems random https://github.com/47deg/sbt-microsites/issues/70
- I had to revert the main scala version to 2.11 because sbt-micro-site is not released for 2.12 https://github.com/47deg/sbt-microsites/issues/82
- colour and logos are not specific to monocle but since I have 0 artistic talent, I would prefer to delegate to someone else :) It is also related to https://github.com/julien-truffaut/Monocle/issues/375


juanpedromoreno:
https://github.com/47deg/sbt-microsites/issues/70 is now fixed after https://github.com/47deg/sbt-microsites/releases/tag/v0.3.1 has been released ;)

julien-truffaut:
I upgraded sbt-microsite which fixes the two main issues: syntax highlighting and scala 2.12

julien-truffaut:
I reviewed the website format such as we get the equivalent of the README in the front page and then we use doc layout for the rest of the resources.

I know that @NightRa preferred that we kept a descriptive README but it makes the snippet likely to be incorrect as we cannot use tut (see https://github.com/julien-truffaut/Monocle/issues/382). 

Except if you feel this is going completely in the wrong direction, I would like to merge and user test the website with  `sbt-microsites`. If they are some fundamental issues, we can always rollback.

julien-truffaut:
I publish from this branch to the website for people to get a better opinion:
http://julien-truffaut.github.io/Monocle/

julien-truffaut:
Going to merge, I think it is a good improvement over the existing solution and I would rather not maintain the existing ad-hoc solution which was a copy-paste from cats.

xuwei-k:
Should we remove sbt-ghpages?

julien-truffaut:
good point, yes we can remove it. It comes with sbt-microsites

PR #412, opened 2016-11-05
Update sbt plugins

jhnsmth:
This PR updates most of the plugins to the latest versions. sbt-site isn't updated, since you're migrating to sbt-microsite.


julien-truffaut:
Thanks! 👍

PR #413, opened 2016-11-06
improve performance of index List and Stream instances

julien-truffaut:
I quickly benchmark it and it seems to improve update performance by at least 50% when the index is within the list.


aoiroaoino:
awesome!! 👍

PR #414, opened 2016-11-10
Extend peripheral modules

jeslg:
This PR contains some changes in the peripheral modules:

* It includes new methods (`extract`, `extracts`, `assign_` and `mod_`) for the existing lens and optional representatives in the state module.
* It adds new representatives in the state module, such as `Getter` and `Setter`.
* It adds a new *reader* module, which includes a `Getter` representative as example.

In general, these ideas have emerged while writing a blog post, ["Lens, State Is Your Father"](https://blog.hablapps.com/2016/11/10/lens-state-is-your-father/), where they are described in detail.


julien-truffaut:
awesome, would you mind to squash your commits? I need to add this instruction somewhere ...

PR #416, opened 2016-11-11
fix warning

xuwei-k:
https://github.com/typesafehub/migration-manager/blob/0.1.11/sbtplugin/src/main/scala/com/typesafe/tools/mima/plugin/Keys.scala#L10-L11


julien-truffaut:
👍

PR #425, opened 2016-11-24
[Non binary compatible change] Review lookup API

julien-truffaut:
- add several methods to lookup target(s) (isEmpty, nonEmpty, find, exist, all, lastOption)
- change the API of a few methods from Traversal and Fold such as it is more consise to use in some scenario, e.g. 
before:
`def exist(p: A => Boolean)(s: S) => Boolean`
after:
`def exist(p: A => Boolean): S => Boolean`
- rename `isMatching` to `nonEmpty` for consistency


julien-truffaut:
I am going to merge as this PR was opened almost 3 weeks ago without review.

PR #429, opened 2016-12-03
Create Optional examples

mdulac:
Add some examples : it already exist for Prism and Iso.

### DIFF
--- a/example/src/main/scala/monocle/example/OptionalExample.scala
+++ b/example/src/main/scala/monocle/example/OptionalExample.scala
@@ -1,0 +1,28 @@
+package monocle.example
+
+import monocle.Optional
+
+object OptionalExample {
+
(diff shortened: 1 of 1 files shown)


example/src/main/scala/monocle/example/OptionalExample.scala
```diff
@@ -0,0 +1,28 @@
+package monocle.example
+
+import monocle.Optional
+
+object OptionalExample {
+
+  def nil[A] = Optional[List[A], Unit] {
```
julien-truffaut: `nil` could be a Prism

julien-truffaut:
@mdulac these examples are mainly there because of a [https://issues.scala-lang.org/browse/SI-7139](SI-7139) which is going to be fix in scala 2.12.1. Once it is done, I would like to move almost all examples to the website.

Sorry for the lack of documentation :(

PR #432, opened 2016-12-04
fix #428 correct Iso javadoc

### LINKED ISSUE #428: Iso documentation bug
The labels on the diagrams are wrong:
```
              get                           reverse.get
     -------------------->             -------------------->
   S                       A         T                       B
     <--------------------             <--------------------
       reverse.reverseGet                   reverseGet
```
For an `Iso[S,T,A,B]`:
`get: S => A` "1st type to 3rd type"
`reverseGet: B => T` "4th type to 2nd type"
`reverse: Iso[B,A,T,S]`

The type of `reverse.reverseGet` is therefore `S => A`, and the type of `reverse.get` is `B => T`.  The arrows on the diagram shouldn't be going in opposite directions.  The implementations of `Iso.reverse` and `IsoLaws` bear this out.

I don't think it's possible to produce an actual isomorphism from a polymorphic `PIso` without an intermediating `A <=> B` or `S <=> T`, ie with a `modify` call in there somewhere.

julien-truffaut:
\cc @kenbot

### DIFF
--- a/core/shared/src/main/scala/monocle/Iso.scala
+++ b/core/shared/src/main/scala/monocle/Iso.scala
@@ -3,34 +3,34 @@ package monocle
 import scalaz.Isomorphism.{<=>, <~>}
 import scalaz.Leibniz.===
 import scalaz.Liskov.<~<
-import scalaz.{Applicative, Category, Functor, Maybe, Monoid, Split, \/}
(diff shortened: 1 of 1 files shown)


julien-truffaut:
Merging as is it is a doc change

PR #434, opened 2016-12-05
Add .lens macro for ApplyLens syntax

cvogt:


macro/shared/src/main/scala/monocle/macros/internal/Macro.scala
```diff
@@ -77,6 +77,7 @@ private[macros] class MacroImpl(val c: blackbox.Context) {
     c.Expr[PLens[S, T, A, B]](q"""
       import monocle.PLens
       import scalaz.Functor
+      import _root_.scala.language.higherKinds // prevent warning at call site
```
julien-truffaut: good idea

julien-truffaut:
LGTM 👍

PR #435, opened 2016-12-06
fix doc issues related to SI-7139 and review doc

julien-truffaut:
- upgrade to scala 2.12.1
- fix doc issues related to SI-7139
- change build to run tut for 2.12 only
- revisit Lens, Iso, Prism and index page


julien-truffaut:
Any idea how not to build the docs module for scala 2.11?

julien-truffaut:
Merging as it is mostly a doc change

PR #442, opened 2016-12-10
Add missing methods for `ApplyFold` and `ApplyPrism`

### LINKED ISSUE #439: Add missing methods in Apply optics
We should go through https://github.com/julien-truffaut/Monocle/blob/master/core/shared/src/main/scala/monocle/syntax/Apply.scala and add all missing methods. It is probably a good idea to fork from https://github.com/julien-truffaut/Monocle/pull/425

kubum:
Fixes #439


julien-truffaut:
👍 Cheers

PR #443, opened 2016-12-10
Adding missing methods to Apply optics - Part2

emmedema:
Fixes part of #439 

Added methods for Iso, Lens, Optional, Traversal, Getter


core/shared/src/main/scala/monocle/syntax/Apply.scala
```diff
@@ -241,10 +254,17 @@ final case class ApplySetter[S, T, A, B](s: S, setter: PSetter[S, T, A, B]) {
 final case class ApplyTraversal[S, T, A, B](s: S, traversal: PTraversal[S, T, A, B]){
   @inline def getAll: List[A] = traversal.getAll(s)
   @inline def headOption: Option[A] = traversal.headOption(s)
+  @inline def LastOption: Option[A] = traversal.lastOption(s)
```
julien-truffaut: would you mind to lower case it

julien-truffaut:
👍 Thanks

PR #446, opened 2016-12-13
Add support for refined Char (lowercase and uppercase)

sindoudou:
Add Prism for refined Char


julien-truffaut:
👍 Thanks @sindoudou

PR #448, opened 2016-12-15
prepare 1.4.0-M1 release

julien-truffaut:
The main goal of this milestone release is to let users test the refactor of typeclass https://github.com/julien-truffaut/Monocle/pull/405.

I will release tomorrow if everything is ok.


julien-truffaut:
release is done

PR #453, opened 2016-12-16
Add .zip to Getter

lunaryorn:
As per [Gitter](https://gitter.im/julien-truffaut/Monocle?at=5853cf11c895451b75e2a6f9) this is `.zip` for Getter.  

@julien-truffaut I looked at `.zip` for `Fold` as well, but I must admit that the implementation wasn't immediately obvious to me, and I didn't have the time to think about it yet.  I'll get back to it when I've got more time.  Would appreciate pointers in the right direction :blush:

### DIFF
--- a/core/shared/src/main/scala/monocle/Getter.scala
+++ b/core/shared/src/main/scala/monocle/Getter.scala
@@ -1,6 +1,6 @@
 package monocle
 
-import scalaz.{Arrow, Choice, Monoid, \/}
+import scalaz.{Arrow, Choice, Monoid, Zip, \/}
 
 /**
  * A [[Getter]] can be seen as a glorified get method between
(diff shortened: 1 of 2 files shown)


julien-truffaut:
Awesome, could you also add an instance of `Zip` for `Getter`. Similar to https://github.com/lunaryorn/Monocle/blob/14f7864d229dc186eb71224a7a5eba3eaa309aa1/core/shared/src/main/scala/monocle/Getter.scala#L145

lunaryorn:
@julien-truffaut I added a Zip instance.  Not sure whether it's correct though, would appreciate a careful review :blush:  

Is there any place where I should add tests for these?

julien-truffaut:
yes you can add a unit test like (just realised about typo in test) https://github.com/julien-truffaut/Monocle/blob/master/test/shared/src/test/scala/monocle/GetterSpec.scala#L24-L26 

When we will move to cats, we will reuse cats laws

lunaryorn:
@julien-truffaut I pushed a test case.

julien-truffaut:
Great, thanks @lunaryorn 

👍

Issue #455, opened 2016-12-16
add unzip for Getter


julien-truffaut:
similar to #453 
see also https://github.com/scalaz/scalaz/blob/series/7.3.x/core/src/main/scala/scalaz/Unzip.scala#L10

PR #458, opened 2016-12-16
Add unzip instance for Getter

lunaryorn:
Closes GH-455

@julien-truffaut While I was at it :)

### DIFF
--- a/test/shared/src/test/scala/monocle/GetterSpec.scala
+++ b/test/shared/src/test/scala/monocle/GetterSpec.scala
@@ -43,6 +43,13 @@ class GetterSpec extends MonocleSuite {
     Zip[Getter[String, ?]].zip(length, upper).get("helloworld") shouldEqual((10, "HELLOWORLD"))
   }
 
+  test("Getter has a Unzip instance") {
+    val lengthAndUpper = Getter[String, (Int, String)](s => s.length -> s.toUpperCase)
+    val (length, upper) = Unzip[Getter[String, ?]].unzip(lengthAndUpper)
+    length.get("helloworld") shouldEqual 10
+    upper.get("helloworld") shouldEqual "HELLOWORLD"
+  }
+
   test("get") {
     i.get(Bar(5)) shouldEqual 5
   }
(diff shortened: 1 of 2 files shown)


julien-truffaut:
if you get the time, could you also add `unzip` to the `Getter` class, the signature should probably look like this:

```
import scalaz.Leibniz.===
def unzip[B, C](implicit ev: (B, C) === A): (Getter[S, B], Getter[S, C])
```

but we should check that type inference works with something like:
```
getterA.zip(getterB).unzip
```

see also https://github.com/julien-truffaut/Monocle/blob/master/core/shared/src/main/scala/monocle/Prism.scala#L246

lunaryorn:
I added the instance method to `Getter`, but it doesn't look like type inference can cope with it.  The compiler won't find the implicit evidence for `===` unless you explicitly give type parameters to `unzip`.

test/shared/src/test/scala/monocle/GetterSpec.scala
```diff
@@ -57,4 +64,26 @@ class GetterSpec extends MonocleSuite {
...
+    length.get("helloworld") shouldEqual 10
+    upper.get("helloworld") shouldEqual "HELLOWORLD"
+
+  }
+
+  test("zip/unzip roundtrip") {
+    val length = Getter[String, Int](_.length)
+    val upper = Getter[String, String](_.toUpperCase)
+
+    val (length1, upper1) = length.zip(upper).unzip[Int, String]
```
lunaryorn: I had to give the type parameters to `.unzip` explicitly to make the compiler find the implicit evidence for `===`.  Without explicit type parameters it wouldn't compile.

julien-truffaut: arf ok, maybe we should revert back to your first commit if we can't get the type inference to work properly

lunaryorn:
@julien-truffaut I dropped the `.unzip` instance method again.

julien-truffaut:
@lunaryorn sorry for the back and force. Thank you very much 👍

lunaryorn:
@julien-truffaut Never mind; I learned a bit about scalaz along the way for I never actually used it before.

PR #460, opened 2016-12-17
Simplify index.html and add basic sbt instructions

### LINKED ISSUE #420: Clearer module documentation
I have met a few people who thought all modules where required to use monocle. I think we should propose a lightweight sbt config containing only `core` and `macro` modules and then explain that other modules are here for particular use cases.

julien-truffaut:
This update greatly simplifies the front page, I got inspired from https://circe.github.io/circe/

fix #420 

I am still not happy about the one liner to describe optics: `Optics are a group of purely functional ...`. It would be great to get some proposal


julien-truffaut:
Doc update

Issue #450, opened 2016-12-15
Feature request:  `Each` but for `Optional`s


kenbot:
In the same way that an `Each` instance provides a canonical traversal for a type, I feel like there's good motivation for a typeclass that provides a canonical `Optional` for a type. I have no idea what to call it.

Obvious instances: `scala.Option`, `scala.Either`, `scala.util.Try`, `scalaz.\/`, `scalaz.Validation`, `cats.Xor`.
It's also very common in userland to have ADTs where one branch contains the payload, and other branches describe various failure conditions.

julien-truffaut:
Good idea, there should probably be a law that says:
`theOne.asTraversal == each` where `theOne` is a bad name for new typeclass optic

kenbot:
How about "possible", as in "a possible int", "a possible banana":
```
abstract class Possible[S,A] {
  def possible: Optional[S,A]
}

trait PossibleFunctions {
  def possible[S, A](implicit ev: Possible[S, A]): Optional[S, A] = ev.possible
}
```

julien-truffaut:
why not? It would be nice to be as concise as possible though

julien-truffaut:
@kenbot feel free to make a PR to add `Possible`, me can always rename it later if we have better option

kenbot:
Further discussion points: 
- The more I think about it, the less I like the name "Possible" -- it seems to describe the optional target value, rather than the source object which contains it. Something like `Wrap`, `Select` or `Focus` would be more in the right spirit, although I don't much like them either.
- I'll raise another PR for `Try` instances in `std`, if you're happy.
- As per the points raised in the PR -- perhaps there should be a corresponding `Each` instance for each `Possible` instance?  Probably using an `asEach` method on `Possible`.

julien-truffaut:
@kenbot I just thought about a use case for `possible` as an `Optional`:

```scala
sealed trait Foo[A]
case class Bar(value: A) extends Foo[A]
case class Baz(value: A) extends Foo[A]
case class Buzz[A]() extends Foo[A]
```

We can implement an `Optional` which zooms into `A` inside of both `Bar` or `Baz`. However, a `Prism` requires to chose between `Bar` or `Baz`. So if `possible` is a `Prism`, it means that `possible` will not always be consistent with `each`. WDYT?

kenbot:
Nice find! I like the idea of being able to support this usecase.  

[...]

PR #461, opened 2016-12-19
New `Possible` typeclass that defines a canonical Optional over a type

kenbot:
As discussed in #450.

The code is basically a facsimile of the `Each` code, with `Optional` instead of `Traversal`, and instances for `Option`, `Either`, `\/`, `Validation` and `Maybe`.  I've omitted support for `scala.util.Try`, which should probably have its own file in `std`, and would be better presented as a separate PR.

The property based test follows `Each` in using the `fromIso` method as a vehicle for hitching the concept to the `Optional` laws.  

I haven't added anything special for `possible.asTraversable == each`, because it concerns types that have both `Each` and `Possible` instances, which only `Maybe` currently satisfies, and possibly warrants further conversation.   

It would be technically correct to make `Possible` a subclass of `Each` (ie every `Possible` yields a `Traversal`, but not every `Each` yields an `Optional`), but Monocle has consistently avoided subclassing, and there is no reason to change policy here. We could, however, provide `Each` instances for each of the `Possible` instances, yielding `Traversal`s that obey the `Optional` laws.

### DIFF
--- a/core/shared/src/main/scala/monocle/std/Try.scala
+++ b/core/shared/src/main/scala/monocle/std/Try.scala
@@ -1,0 +1,27 @@
+package monocle.std
+
+import monocle.{Prism, PPrism}
+
+import scala.util.{Try, Success, Failure}
+
+import scalaz.{\/, -\/, \/-}
+
+object utilTry extends TryOptics
+
+trait TryOptics {
+
+  final def pTrySuccess[A, B]: PPrism[Try[A], Try[B], A, B] =
+    PPrism[Try[A], Try[B], A, B] {
+      case Success(a) => \/-(a) 
+      case Failure(e) => -\/(Failure(e)) 
+    }(Success.apply)
+
+  final def trySuccess[A]: Prism[Try[A], A] =
+    pTrySuccess[A, A]
+
+  final def tryFailure[A]: Prism[Try[A], Throwable] =
+    Prism[Try[A], Throwable] {
+      case Success(a) => None
+      case Failure(e) => Some(e)
+    }(Failure.apply)
+}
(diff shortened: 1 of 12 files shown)


julien-truffaut:
We should add tests for each instance of `Possible`, e.g. in `OptionSpec`

I agree we don't need to add more complex tests or helpers in this PR but I think `Try` instance should be in the companion object of `Possible`.

I wonder if we shouldn't make `possible` a `Prism` instead of an `Optional`? It would create a polymorphic constructor for `success`

julien-truffaut:
You're right, the only thing an `Optional` would allow is an instance for something like:

```scala
sealed trait Foo[A]
case class Bar[A](value: A, id: Long) extends Foo[A] // and eventually other fields
case class Fizz[A]() extends Foo[A]

object Foo {
  implicit def possible[A]: Possible[Foo[A], A] = ??? // if Optional
  implicit def possible[A]: Possible[Foo[A], Bar[A]] = ??? // if Prism
}
```

I think it might be better to be more strict since we don't foresee any real use case where `Optional` is better. If such use case happens later, we can always loosen then definition of `possible` or create a new one.

core/shared/src/main/scala/monocle/std/Try.scala
```diff
@@ -0,0 +1,27 @@
...
+
+import monocle.{Prism, PPrism}
+
+import scala.util.{Try, Success, Failure}
+
+import scalaz.{\/, -\/, \/-}
+
+object utilTry extends TryOptics
+
+trait TryOptics {
```
julien-truffaut: cool, I thought we already had optics for `Try`. Good catch!

julien-truffaut:
It is great, thanks! Would you mind to squash your commits before I merge?

julien-truffaut:
Thanks @kenbot 👍

PR #463, opened 2016-12-19
remove scalac options to compile scala 2.11 with java8

### LINKED ISSUE #462: Support java 7 with scala 2.11
I have had a compilation error trying to use Monocle 1.3.2 in a Java 7 environment. Please build monocle for Java <8 with Scala 2.11 if you can do so without sacrificing functionality.

Yes, Oracle Java 7 is EOL, I know, but I am working on several projects which target the IBM JVM, where 7 is still very much the norm.

julien-truffaut:
Fix #462


julien-truffaut:
Thanks @xuwei-k

PR #467, opened 2016-12-22
Fix document

n4to4:
I read through the document and found these.
(My fixes can be wrong, so please someone review carefully :)


julien-truffaut:
Thank you very much @n4to4 for the careful review. It is extremely useful!

julien-truffaut:
👍

PR #468, opened 2016-12-22
Change Possible from Prism to Optional and add link in website

### LINKED ISSUE #450: Feature request:  `Each` but for `Optional`s

julien-truffaut:
fix #450

### DIFF
--- a/core/shared/src/main/scala/monocle/function/Possible.scala
+++ b/core/shared/src/main/scala/monocle/function/Possible.scala
@@ -1,59 +1,62 @@
 package monocle.function
 
-import monocle.{Iso, Prism}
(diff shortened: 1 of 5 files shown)


julien-truffaut:
Merging as @kenbot is the most familiar with this code

PR #469, opened 2016-12-22
Prepare 1.4.0-M2

julien-truffaut:
should be merged after https://github.com/julien-truffaut/Monocle/pull/468


julien-truffaut:
released

PR #474, opened 2016-12-27
Add Unzip instance for Lens, Fold, Optional, Traversal, Setter

### LINKED ISSUE #464: Add Unzip instance for Lens, Fold, Optional, Traversal, Setter
see https://github.com/julien-truffaut/Monocle/issues/455

We can use `first` and `second` instance from `Tuple2` (https://github.com/julien-truffaut/Monocle/blob/master/example/src/test/scala/monocle/function/FieldsExample.scala)

### LINKED ISSUE #473: Add Each instance for Either, Disjunction and Validation

n4to4:
Fix #464 hopefully :)

### DIFF
--- a/core/shared/src/main/scala/monocle/Lens.scala
+++ b/core/shared/src/main/scala/monocle/Lens.scala
@@ -1,6 +1,6 @@
 package monocle
 
-import monocle.function.fields._
+import monocle.function.fields.{first, second}
 import scalaz.{Applicative, Choice, Functor, Monoid, Split, Unzip, \/}
 
 /**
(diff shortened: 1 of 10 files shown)


core/shared/src/main/scala/monocle/Lens.scala
```diff
@@ -1,6 +1,7 @@
 package monocle
 
-import scalaz.{Applicative, Choice, Functor, Monoid, Split, \/}
+import monocle.function.fields._
```
julien-truffaut: it is a tiny details but would you mind to specify which optics do you import. I think it would help the reader: `import monocle.function.fields.{first, second}`

n4to4:
@julien-truffaut Done :)

julien-truffaut:
👍 thanks a lot

PR #480, opened 2016-12-30
Add Support for refined startsWith. Fixes #477

### LINKED ISSUE #477: Add startsWith Prism
`def startsWith(prefix: String): Prism[String, String Refined StartsWith[prefix.type]]`

see also #446 and #449

valydia:


build.sbt
```diff
@@ -134,7 +135,7 @@ lazy val lawJS  = law.js
 lazy val law    = crossProject.dependsOn(core)
   .settings(moduleName := "monocle-law")
   .configureCross(monocleCrossSettings)
-  .settings(libraryDependencies ++= Seq(discipline.value, scalacheck.value))
+  .settings(libraryDependencies ++= Seq(discipline.value, scalacheck.value, refinedScalacheck.value))
```
julien-truffaut: is it needed for law or for test?

refined/shared/src/main/scala/monocle/refined/startswith.scala
```diff
@@ -0,0 +1,17 @@
+package monocle.refined
+
+import eu.timepit.refined.api.Refined
+import eu.timepit.refined.string.StartsWith
+import monocle._
+
+object startswith extends StartsWithInstances
```
julien-truffaut: Could you rename it to `string` and `StringInstances`?

build.sbt
valydia: For test only actually! I'll move it in the relevant section.

refined/shared/src/main/scala/monocle/refined/startswith.scala
valydia: To be consistant with the other refined Instances (`CharsInstances`, `BitsInstances`), I've renamed it to `StringsInstances` (instead of `StringInstances`)! But I can happily change it to `StringInstances` if you prefer :-).

julien-truffaut: Perfect thanks

julien-truffaut:
👍 Thank you!

refined/shared/src/main/scala/monocle/refined/strings.scala
```diff
@@ -0,0 +1,17 @@
...
+import eu.timepit.refined.string.StartsWith
+import monocle._
+
+object strings extends StringsInstances
+
+trait StringsInstances {
+
+  def startsWith(prefix: String): Prism[String, StartsWithString[prefix.type]] = {
+    Prism.partial[String, Refined[String, StartsWith[prefix.type]]] {
+      case string if string.startsWith(prefix) => Refined.unsafeApply(string)
```
fthomas: refined already knows how to check if a `String` satisfies the `StartsWith[prefix.type]` predicate via its `Validate` type class. If the `startsWith` method would take an `implicit v: Validate[String, StartsWith[prefix.type]]` the check `string.startsWith(prefix)` could be replaced with `v.isValid(prefix)`. This would be my first step to generalize the refined `Prism` definitions.

valydia: Yes indeed! I'll submit another PR! Thanks for the feedback

test/shared/src/test/scala/monocle/refined/StringsSpec.scala
```diff
@@ -0,0 +1,23 @@
...
+import eu.timepit.refined.scalacheck.string.startsWithArbitrary
+import eu.timepit.refined._
+
+
+import scalaz.Equal
+
+
+class StringsSpec extends MonocleSuite {
+
+  implicit val startsWithCoGen: Cogen[StartsWithString[W.`"hello"`.T]] = Cogen[String].contramap[StartsWithString[W.`"hello"`.T]](_.value)
```
fthomas: I think this could be added to refined-scalacheck for all refined types as:
```scala
implicit def refTypeCogen[F[_, _], T: Cogen, P](implicit rt: RefType[F]): Cogen[F[T, P]] =
    Cogen[T].contramap(tp => rt.unwrap(tp))
```
`RefType` is a type class that abstracts over `Refined` and other tag types so we wouldn't have to define `Cogen` instances for specific refined types anymore.

PR #482, opened 2016-12-30
Improve StartsWith Prism using refined Validate

valydia:
Improvement of the PR #480 based on @fthomas's feedback.

### DIFF
--- a/refined/shared/src/main/scala/monocle/refined/strings.scala
+++ b/refined/shared/src/main/scala/monocle/refined/strings.scala
@@ -1,6 +1,6 @@
 package monocle.refined
 
-import eu.timepit.refined.api.Refined
+import eu.timepit.refined.api.{Validate, Refined}
 import eu.timepit.refined.string.StartsWith
 import monocle._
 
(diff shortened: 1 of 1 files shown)


refined/shared/src/main/scala/monocle/refined/strings.scala
```diff
@@ -1,16 +1,16 @@
...
+import eu.timepit.refined.api.{Validate, Refined}
 import eu.timepit.refined.string.StartsWith
 import monocle._
 
 object strings extends StringsInstances
 
 trait StringsInstances {
 
-  def startsWith(prefix: String): Prism[String, StartsWithString[prefix.type]] = {
+  def startsWith(prefix: String)(implicit v: Validate[String, StartsWith[prefix.type]]): Prism[String, StartsWithString[prefix.type]] = {
```
julien-truffaut: just to be sure I understand correctly, we cannot fetch a `Validate[String, StartsWith[prefix.type]` inside the method definition because we need know `prefix`?

fthomas: scalac will complain with "could not find implicit value for parameter v: Validate" if we do not specify it as implicit parameter. The reason is that the implicit which provides the `Validate[String, StartsWith[S]]` instance is a `def` which itself requires an implicit `shapeless.Witness[S]`. If you also fetch a `Witness[prefix.type]` in the method definition and make it implicit, you could also fetch `Validate[String, StartsWith[prefix.type]]` in the method without the need to specify as implicit parameter. The advantage of specifying it as implicit parameter is that scalac will fill in all required implicits automatically and that you don't need to know all the dependencies of the `Validate[String, StartsWith[prefix.type]]` instance.

refined/shared/src/main/scala/monocle/refined/strings.scala
```diff
@@ -1,16 +1,16 @@
...
     Prism.partial[String, Refined[String, StartsWith[prefix.type]]] {
-      case string if string.startsWith(prefix) => Refined.unsafeApply(string)
+      case string if v.isValid(prefix) => Refined.unsafeApply(string)
```
fthomas: Just noticed that this should be `v.isValid(string)` since `prefix` always satisfies `StartsWith[prefix.type]`. :-)

julien-truffaut:
A similar change was made in #483. Thank you @valydia

Issue #481, opened 2016-12-30
Add endsWith Prism


valydia:
Similarly to #477, we could implements the endsWith Prism:

```def endsWith(suffix: String): Prism[String, String Refined EndsWith[suffix.type]]```

PR #483, opened 2017-01-01
Add Support for refined endsWith #481

mrcmatuszak:
Basically it's similar to one from 'startsWith' Prism with fthomas comments applied.

### DIFF
--- a/refined/shared/src/main/scala/monocle/refined/strings.scala
+++ b/refined/shared/src/main/scala/monocle/refined/strings.scala
@@ -8,16 +8,17 @@ object strings extends StringsInstances
 
 trait StringsInstances {
 
-  def startsWith(prefix: String): Prism[String, StartsWithString[prefix.type]] = {
-    Prism.partial[String, Refined[String, StartsWith[prefix.type]]] {
-      case string if string.startsWith(prefix) => Refined.unsafeApply(string)
-    }{_.value}
-  }
+  def startsWith(prefix: String)(implicit v: Validate[String, StartsWith[prefix.type]]): Prism[String, StartsWithString[prefix.type]] =
+    refinedPrism[String, StartsWith[prefix.type]](prefix)
 
-  def endsWith(suffix: String)(implicit v: Validate[String, EndsWith[suffix.type]]): Prism[String, EndsWithString[suffix.type]] = {
-    Prism.partial[String, Refined[String, EndsWith[suffix.type]]] {
-      case string if v.isValid(string) => Refined.unsafeApply(string)
-    }{_.value}
-  }
+  def endsWith(suffix: String)(implicit v: Validate[String, EndsWith[suffix.type]]): Prism[String, EndsWithString[suffix.type]] =
+    refinedPrism[String, EndsWith[suffix.type]](suffix)
 
+  private def refinedPrism[T, P](t: T)(implicit v: Validate[T, P]): Prism[T, T Refined P] = {
+    Prism.partial[T, Refined[T, P]] {
+      case tt if v.isValid(t) => Refined.unsafeApply[T, P](tt)
+    } {
+      _.value
+    }
+  }
 }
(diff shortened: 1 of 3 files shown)


fthomas:
LGTM! If you want to go the next step of generalizing the `Prism` definitions, you could factor out a `def refinedPrism[T, P](implicit v: Validate[T, P]): Prism[T, T Refined P] = ...` and implement all four refined prisms with this function.

mrcmatuszak:
Add rest of refined functions: MatchesRegex, Regex, Uri, Url, UUID, Xml, Xpath?

refined/shared/src/main/scala/monocle/refined/strings.scala
```diff
@@ -8,16 +8,17 @@ object strings extends StringsInstances
...
-    Prism.partial[String, Refined[String, EndsWith[suffix.type]]] {
-      case string if v.isValid(string) => Refined.unsafeApply(string)
-    }{_.value}
-  }
+  def endsWith(suffix: String)(implicit v: Validate[String, EndsWith[suffix.type]]): Prism[String, EndsWithString[suffix.type]] =
+    refinedPrism[String, EndsWith[suffix.type]](suffix)
 
+  private def refinedPrism[T, P](t: T)(implicit v: Validate[T, P]): Prism[T, T Refined P] = {
+    Prism.partial[T, Refined[T, P]] {
+      case tt if v.isValid(t) => Refined.unsafeApply[T, P](tt)
```
fthomas: We need to check if `tt` is valid since this is the value we are wrapping with `Refined`. The `t` parameter here will be `prefix` and `suffix` from above and therefore `v.isValid(t)` will always be true. The `t` parameter is not needed to define `refinedPrism`. If you remove that parameter, you can also use `refinedPrism` to implement `chars.lowerCase` and `chars.upperCase` with it.

refined/shared/src/main/scala/monocle/refined/strings.scala
```diff
@@ -8,16 +8,17 @@ object strings extends StringsInstances
 
 trait StringsInstances {
 
-  def startsWith(prefix: String): Prism[String, StartsWithString[prefix.type]] = {
-    Prism.partial[String, Refined[String, StartsWith[prefix.type]]] {
-      case string if string.startsWith(prefix) => Refined.unsafeApply(string)
-    }{_.value}
-  }
+  def startsWith(prefix: String)(implicit v: Validate[String, StartsWith[prefix.type]]): Prism[String, StartsWithString[prefix.type]] =
+    refinedPrism[String, StartsWith[prefix.type]](prefix)
```
fthomas: I think you can omit the type parameters to `refinedPrism` here, type inference will figure them out.

refined/shared/src/main/scala/monocle/refined/strings.scala
mrcmatuszak: of course it should be `tt` not `t`, lame mistake.

refined/shared/src/main/scala/monocle/refined/strings.scala
```diff
@@ -1,17 +1,24 @@
...
+
+  def endsWith(suffix: String)(implicit v: Validate[String, EndsWith[suffix.type]]): Prism[String, EndsWithString[suffix.type]] =
+    refinedPrism[String, EndsWith[suffix.type]](suffix)
 
+  private def refinedPrism[T, P](t: T)(implicit v: Validate[T, P]): Prism[T, T Refined P] = {
```
fthomas: Btw, if we make this public, people can use it for their own custom predicates that are not in refined.

julien-truffaut: Yes I think we should expose a method like that but in a more general package (I don't think it is specific to Refined String). We can do that on another PR.

julien-truffaut:
👍 Thanks @mrcmatuszak and @fthomas for the review

PR #486, opened 2017-01-02
Use refinedPrism to implement chars prisms

fthomas:
This uses `refinedPrism` (which was added in #483) to implement
`chars.lowerCase` and `chars.upperCase`.


julien-truffaut:
Thanks 👍

PR #487, opened 2017-01-02
Add UUID Prism. Fixes #485

### LINKED ISSUE #485: Add uuid Prism
```
val uuid: Prism[String, UUID]
```

We can add it in core.std.string as I don't think we need `refined`

mjjc:


test/shared/src/test/scala/monocle/std/StringsSpec.scala
```diff
@@ -1,11 +1,29 @@
...
 import monocle.law.discipline.function._
 import monocle.law.discipline.{IsoTests, PrismTests, TraversalTests}
+import org.scalacheck.Arbitrary
+
+import scalaz.Equal
 
 class StringsSpec extends MonocleSuite {
+  implicit def arbitraryUUID: Arbitrary[UUID] = Arbitrary(UUID.randomUUID)
+
+  implicit def equalUUID: Equal[UUID] = new Equal[UUID] {
```
julien-truffaut: You can use `equalA` https://github.com/scalaz/scalaz/blob/series/7.3.x/core/src/main/scala/scalaz/Equal.scala#L44

test/shared/src/test/scala/monocle/std/StringsSpec.scala
```diff
@@ -1,11 +1,29 @@
...
+    override def equal(a1: UUID, a2: UUID): Boolean = a1 == a2
+  }
+
+  implicit def arbitraryUUIDtoUUID: Arbitrary[UUID => UUID] = Arbitrary(
```
julien-truffaut: it should be enough to have a `Cogen[UUID]`. Btw, you can move this definition to `TestInstances`

mjjc:
Thanks for the hints!

julien-truffaut:
👍 Thanks

PR #495, opened 2017-01-13
set sudo: true

xuwei-k:
```yaml
env:
  sudo: false
```

is no effect. Right way is

```yaml
sudo: false
```

but if set `sudo: false` scala-js tests failure.


n4to4:
> if set sudo: false scala-js tests failure.

This has been occurring on my CI too, and this PR seems to fix it 😄

julien-truffaut:
Cheers 👍

PR #499, opened 2017-01-21
make GenLens macro work for private[package] members

### LINKED ISSUE #498: GenLens for case classes with private[package] members doesn't work
Consider the following example

```scala
object A {
  case class B(private[A] val c: Int, private[A] val d: Int)

  def lensOfC: Lens[B, Int] = GenLens[B](_.c)
}
```

This gives the error message `Cannot find method c in A`. But `c` is perfectly visible in the place where `GenLens` is called, so in my view this should be possible.

The same problem occurs when using `private[B] val c` and using it in the companion object `B`.

fkz:
this fixes #498

### DIFF
--- a/macro/shared/src/main/scala/monocle/macros/internal/Macro.scala
+++ b/macro/shared/src/main/scala/monocle/macros/internal/Macro.scala
@@ -60,8 +60,14 @@ private[macros] class MacroImpl(val c: blackbox.Context) {
 
     val strFieldName = c.eval(c.Expr[String](c.untypecheck(fieldName.tree.duplicate)))
 
+    // when using 'private[package] val name' members, they get mangled to 'name$access$number' by the scala compiler
+    // so also accept these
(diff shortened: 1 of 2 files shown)


macro/shared/src/main/scala/monocle/macros/internal/Macro.scala
```diff
@@ -60,8 +60,14 @@ private[macros] class MacroImpl(val c: blackbox.Context) {
 
     val strFieldName = c.eval(c.Expr[String](c.untypecheck(fieldName.tree.duplicate)))
 
+    // when using 'private[package] val name' members, they get mangled to 'name$access$number' by the scala compiler
+    // so also accept these
+    def isNameOrMangledFrom(symbolName: String, fieldName: String): Boolean = {
```
julien-truffaut: could you change the first parameter to be a `c.TermName` instead of `String` such as we call it `isNameOrMangledFrom(m.name, strFieldName)`?

Issue #451, opened 2016-12-15
Add Equal instance for optics in test


julien-truffaut:
optics are functions, so we could approximate an `Equal` instance as it is done in cats:
https://github.com/typelevel/cats/blob/master/laws/src/main/scala/cats/laws/discipline/Eq.scala#L11-L15

julien-truffaut:
done

PR #500, opened 2017-01-22
Add Equal instance for optics in test

mjjc:
As per #451


julien-truffaut:
nice, could you move the tests to another file? `MacroOutSideMonocleSpec` is here to check macro in another package file than monocle.

mjjc:
Right, now it makes more sense.

mjjc:
Previous version was easily extensible to polymorphic optics, so I did that.

julien-truffaut:
Thanks @mjjc 👍

PR #502, opened 2017-02-25
Adds coalgebroid-based constructors for Lens and Traversal

jeslg:
This includes an alternative constructor for lenses based on `Store` coalgebras. In fact, this constructor is even more generic, since it uses the `StoreComonad` typeclass, where `pos` and `peek` can be found.

(*) Similar constructors could be included for `PLens` (`S => IndexedStore[A, B, T]`), `Traversal` (`S => FreeAp[Store[A, ?], S]`) and `PTraversal` (`S => FreeAp[IndexedStore[A, B, T]]`).

### DIFF
--- a/core/shared/src/main/scala/monocle/Lens.scala
+++ b/core/shared/src/main/scala/monocle/Lens.scala
@@ -1,7 +1,7 @@
 package monocle
 
 import monocle.function.fields.{first, second}
-import scalaz.{Applicative, Choice, ComonadStore, Functor, Monoid, Split, Unzip, \/}
+import scalaz.{Applicative, Choice, IndexedStore, Functor, Monoid, Split, Unzip, \/}
 
 /**
  * A [[PLens]] can be seen as a pair of functions:
@@ -224,6 +224,12 @@ object PLens extends LensInstances {
     )(t => _.bimap(_ => t, _ => t))
 
   /**
+   * create a [[PLens]] from a context (indexed store) coalgebra.
+   */
+  def contextCoalg[S, T, A, B](f: S => IndexedStore[A, B, T]): PLens[S, T, A, B] =
+    PLens[S, T, A, B](s => f(s).pos)(b => s => f(s).peek(b))
+
+  /**
    * create a [[PLens]] using a pair of functions: one to get the target, one to set the target.
    * @see macro module for methods generating [[PLens]] with less boiler plate
    */
@@ -250,10 +256,6 @@ object Lens {
 
   def codiagonal[S]: Lens[S \/ S, S] =
     PLens.codiagonal
-
-  /** creates a [[Lens]] from a Store comonad coalgebra */
-  def storeFCoalg[F[_], S, A](f: S => F[S])(implicit CS: ComonadStore[F, A]): Lens[S, A] =
-    Lens[S, A](s => CS.pos(f(s)))(a => s => CS.peek(a, f(s)))
 
   /** alias for [[PLens]] apply with a monomorphic set function */
   def apply[S, A](get: S => A)(set: A => S => S): Lens[S, A] =
--- a/core/shared/src/main/scala/monocle/Lens.scala
+++ b/core/shared/src/main/scala/monocle/Lens.scala
@@ -1,7 +1,7 @@
 package monocle
 
 import monocle.function.fields.{first, second}
-import scalaz.{Applicative, Choice, Functor, Monoid, Split, Unzip, \/}
+import scalaz.{Applicative, Choice, ComonadStore, Functor, Monoid, Split, Unzip, \/}
 
 /**
  * A [[PLens]] can be seen as a pair of functions:
@@ -250,6 +250,10 @@ object Lens {
 
   def codiagonal[S]: Lens[S \/ S, S] =
     PLens.codiagonal
+
+  /** creates a [[Lens]] from a Store comonad coalgebra */
+  def storeFCoalg[F[_], S, A](f: S => F[S])(implicit CS: ComonadStore[F, A]): Lens[S, A] =
+    Lens[S, A](s => CS.pos(f(s)))(a => s => CS.peek(a, f(s)))
 
   /** alias for [[PLens]] apply with a monomorphic set function */
   def apply[S, A](get: S => A)(set: A => S => S): Lens[S, A] =
--- a/core/shared/src/main/scala/monocle/Traversal.scala
+++ b/core/shared/src/main/scala/monocle/Traversal.scala
@@ -229,7 +229,9 @@ object PTraversal extends TraversalInstances {
   def apply[S, T, A, B](f: S => FreeAp[IndexedStore[A, B, ?], T]) =
     new PTraversal[S, T, A, B] {
       def modifyF[F[_]: Applicative](g: A => F[B])(s: S): F[T] =
-        f(s).foldMap(λ[IndexedStore[A, B, ?] ~> F](is => g(is.pos).map(is.set)))
+        f(s).foldMap(new (IndexedStore[A, B, ?] ~> F) {
+          def apply[X](is: IndexedStore[A, B, X]) = g(is.pos).map(is.set)
+        })
     }
 }
 
--- a/core/shared/src/main/scala/monocle/Traversal.scala
+++ b/core/shared/src/main/scala/monocle/Traversal.scala
@@ -5,10 +5,11 @@ import scalaz.Id.Id
 import scalaz.std.anyVal._
 import scalaz.std.list._
 import scalaz.std.option._
+import scalaz.syntax.functor._
 import scalaz.syntax.std.boolean._
 import scalaz.syntax.std.option._
 import scalaz.syntax.tag._
-import scalaz.{Applicative, Choice, Const, Functor, Monoid, Traverse, Unzip, \/}
+import scalaz.{Applicative, Choice, Const, IndexedStore, FreeAp, Functor, Monoid, Traverse, Unzip, \/, ~>}
 
 /**
  * A [[PTraversal]] can be seen as a [[POptional]] generalised to 0 to n targets
@@ -225,6 +226,11 @@ object PTraversal extends TraversalInstances {
         Applicative[F].apply6(f(get1(s)), f(get2(s)), f(get3(s)), f(get4(s)), f(get5(s)), f(get6(s)))(_set(_, _, _, _, _, _, s))
     }
 
+  def apply[S, T, A, B](f: S => FreeAp[IndexedStore[A, B, ?], T]) =
+    new PTraversal[S, T, A, B] {
+      def modifyF[F[_]: Applicative](g: A => F[B])(s: S): F[T] =
+        f(s).foldMap(λ[IndexedStore[A, B, ?] ~> F](is => g(is.pos).map(is.set)))
+    }
 }
 
 object Traversal {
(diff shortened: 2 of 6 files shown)


test/shared/src/test/scala/monocle/LensSpec.scala
```diff
@@ -27,6 +27,9 @@ class LensSpec extends MonocleSuite {
   val s = Lens[Example, String](_.s)(s => ex => ex.copy(s = s))
   val p = Lens[Example, Point](_.p)(p => ex => ex.copy(p = p))
 
+  implicit def ev = IndexedStoreT.storeTComonadStore[Id.Id, String]
```
julien-truffaut: what other instance of `ComandStore` exists we could for lenses?

core/shared/src/main/scala/monocle/Lens.scala
```diff
@@ -251,6 +251,10 @@ object Lens {
   def codiagonal[S]: Lens[S \/ S, S] =
     PLens.codiagonal
 
+  /** creates a [[Lens]] from a Store comonad coalgebra */
+  def storeFCoalg[F[_], S, A](f: S => F[S])(implicit CS: ComonadStore[F, A]): Lens[S, A] =
```
julien-truffaut: does `f` need to satisfy any law to create a lawful `Lens`?

test/shared/src/test/scala/monocle/LensSpec.scala
jeslg: It seems that *scalaz* contains only two instances of `ComonadStore`: `StoreT` and `TracedT`. Honestly, I've never dealt with the second one, but it doesn't appear to be very practical for the case of *Lenses* (for instance, having [`copoint`](https://github.com/scalaz/scalaz/blob/series/7.3.x/core/src/main/scala/scalaz/TracedT.scala#L20) using `Monoid.zero` to feed the inner function seems odd). Given this situation, do you think it would be preferable to work directly with `Store`?

core/shared/src/main/scala/monocle/Lens.scala
jeslg: As long as `f` is a *comonad coalgebra*, classical lens laws should be satisfied (*PutGet*, *GetPut* and *PutPut*).

julien-truffaut: what does it mean that `f: S => F[S]` is a `comonad coalgebra`?

jeslg: It's a *coalgebra* that commutes with the *comonad* operations. The idea can be found [here](https://www.schoolofhaskell.com/user/tel/lenses-from-scratch) and especially [here](https://bartoszmilewski.com/2015/07/13/from-lenses-to-yoneda-embedding/) (*The Algebraic Picture* section). It seems that the *comonad coalgebra* laws have a direct correspondence with the *lens* ones. 

As I said in the description, I've been toying around with the counterpart constructor for *Traversals*, but still don't know if traversal laws hold so nicely in that context (comonad coalgebras where `FreeAp[Store[A, ?], ?]` is the functor).

julien-truffaut: sorry for the delay, I was on holiday last week. I will try to go through the links you posted as fast as possible!

julien-truffaut:
Thanks very much for links @jeslg, in particular I wasn't aware of the relation between `Store` and Van Laarhoven lens.

You PR looks good but I wonder if it is something we will ever want to use. I know that trying to figure out usage is real guess at best but do you see a use case where you would prefer to use `storeFCoalg` instead of other `Lens` constructors?

jeslg:
Hi @julien-truffaut! I find that relation extremely beautiful :smile: In my case, I was working with several *Store-coalgebra* optics when I missed such a constructor. Then I saw `ComonadStore` in *scalaz*, so I thought it would be nice to reflect the connection in *Monocle*. Anyway, I see your point, `storeFCoalg` doesn't add much, at least in the `Lens` context.

However, I've been taking these ideas to Monocle's `PTraversal`, where it could make more sense. In particular, one can find a range of constructors from `apply1` to `apply6`. What if you need more "focuses"? Wouldn't it be nice to have an alternative `applyN` as the one available for simple `Traversal`s? That's exactly what `S => FreeAp[IndexedStore[A, B, ?], T]` is able to encapsulate. Here's my humble implementation:

```scala
def applyN[S, T, A, B](f: S => FreeAp[IndexedStore[A, B, ?], T]) =
  new PTraversal[S, T, A, B] {
    def modifyF[F[_]: Applicative](g: A => F[B])(s: S): F[T] =
      f(s).foldMap(λ[IndexedStore[A, B, ?] ~> F](is => g(is.pos).map(is.set)))
  }
```

If you think that this method does make sense, keeping the one for *Lenses* is reasonable, just to keep *optic construction* uniform. Thanks for your comments!

julien-truffaut:
`S => FreeAp[IndexedStore[A, B, ?], T]` looks really interesting, if you have time it would be nice to incorporate it within this PR with examples on how to use it.

One thing that annoys with `Traversal.applyN` is that it only accepts `Lens` while we could have generalised it to `Optional`. However, since optics do not inherit from each other we would need to use `lens.asOptional` whenever we would use `applyN` with a `Lens`. Will the `S => FreeAp[IndexedStore[A, B, ?], T]` representation permit to mix optional and required fields? e.g.

```
case class Person(name: String, email: Option[String])
```

jeslg:
Hi there, I've been struggling with the compiler, because inference for `FreeAp[IndexedStore[A, B, ?], T]` in Scala is a nightmare, so examples become quite noisy. As a consequence, I've been thinking of introducing new syntax to deal with `FreeAp`s for `IndexedStore`, to make `applyN` practical. I'll let you know about my progress on that issue.

Regarding mixing optional and required fields, I'm afraid this won't help. `S => FreeAp[IndexedStore[A, B, ?], T]` is just a way of encapsulating a (variable) bunch of getters and the associated (big) setter. In fact, it's this unified setter the one that makes this representation different from a bunch of lenses (and therefore suitable for `PTraversal`s). If we ignore that detail, they are pretty much the same, so the mixing problem remains.

jeslg:
@julien-truffaut, here it is the aforementioned syntax. Now, it's possible to define a `Traversal` in this way:
```scala
val ptraversalN: Traversal[ManyPropObject, Int] = PTraversal { obj =>
  (obj.p1 ~ obj.p2 ~ obj.p4 ~ obj.p5 ~ obj.p6 ~ obj.p7 ~ obj.p8) {
    ManyPropObject(_, _, obj.p3, _, _, _, _, _)
  }
}
```
Besides, it's also possible to define a one-focus `Traversal`:
```scala
val ptraversalN: Traversal[ManyPropObject, Int] = PTraversal { obj =>
  obj.p1 { p1 => obj.copy(p1 = p1) }
}
``` 
And what's more interesting, this works for `PTraversal` as well:
```scala
def tuplePTraversal[A, B, C]: PTraversal[(A, A, C), (B, B, C), A, B] = PTraversal {
  case (a1, a2, c) => (a1 ~ a2) { (_, _, c) }
}
```
To do so, I've followed the [scalaz pattern for `|@|`](https://github.com/scalaz/scalaz/blob/fabab8f699d56279d6f2cc28d02cc2b768e314d7/core/src/main/scala/scalaz/syntax/ApplicativeBuilder.scala), boilerplate that could be generated by a macro in a near future. As you know well, this is still very experimental, so any feedback is more than welcome.

julien-truffaut:
it seems that kind-projector syntax is not supported in 2.10. If there is no easy fix, I would not worry too much as I think we should stop supporting 2.10.

I really like that it generalises to `PTraversal` 🎉

jeslg:
I'm very glad you like it! :smile: 

The fix is trivial, but there seem to be other issues with traversal syntax in 2.10. Particularly, compiler is unable to infer `B` and `T` in `toTraversalBuilder[A, B, T]`. Therefore, this (ugly) manual invocation is necessary:
```scala
val ptraversalN: Traversal[ManyPropObject, Int] = PTraversal { obj =>
  (toTraversalBuilder(obj.p1) ~ obj.p2 ~ obj.p4 ~ obj.p5 ~ obj.p6 ~ obj.p7 ~ obj.p8) {
    ManyPropObject(_, _, obj.p3, _, _, _, _, _)
  }
}
```
Any idea or workaround to improve type inference in this context?

julien-truffaut:
@jeslg no sorry I don't know how to help type inference. I am fine if the test use manual invocation as we will need to add some doc on the website which use 2.12.

jeslg:
Ok. Then, here it is!

core/shared/src/main/scala/monocle/Lens.scala
julien-truffaut: what do you think of replacing ComonadStore by IndexedStore? It would make the relation with PTraversal.applyN more obvious.

jeslg: I like `ComonadStore`, since it decouples particular representations, but given the lack of instances for it, I find your suggestion more than reasonable. Thereby, I've replaced this constructor by an `IndexedStore` version and moved it to `PLens`.

core/shared/src/main/scala/monocle/Lens.scala
```diff
@@ -224,6 +224,12 @@ object PLens extends LensInstances {
     )(t => _.bimap(_ => t, _ => t))
 
   /**
+   * create a [[PLens]] from a context (indexed store) coalgebra.
+   */
+  def contextCoalg[S, T, A, B](f: S => IndexedStore[A, B, T]): PLens[S, T, A, B] =
```
julien-truffaut: would `storeCoalgebra` clearer than `contextCoalg`?

core/shared/src/main/scala/monocle/Traversal.scala
```diff
@@ -225,6 +226,13 @@ object PTraversal extends TraversalInstances {
         Applicative[F].apply6(f(get1(s)), f(get2(s)), f(get3(s)), f(get4(s)), f(get5(s)), f(get6(s)))(_set(_, _, _, _, _, _, s))
     }
 
+  def apply[S, T, A, B](f: S => FreeAp[IndexedStore[A, B, ?], T]) =
```
julien-truffaut: shall we give it the same name that the `Lens` constructor?

julien-truffaut:
👍 Thank you very much @jeslg 

I just put a few comments regarding naming. You don't need to update the PR, I can rename it later if we agree on a new names.

core/shared/src/main/scala/monocle/Lens.scala
jeslg: The thing is that after introducing `IndexedStore`, `f` is no longer a *coalgebra* (`S => F[S]`). It becomes something different (`S => F[T]`). There aren't many references about this thing, but at least in the lens context, it seems that the right name for `f` should be [*coalgebroid*](http://r6.ca/blog/20120623T104901Z.html). On the other hand, I know that `IndexedStore` is also known as "context", but I don't know to what extent. Honestly, I don't like `contextCoalg` much, so please, feel free to rename it to whatever you like.

core/shared/src/main/scala/monocle/Traversal.scala
jeslg: I think that using the same name for both would be ok. Eventually, if users adopted `Traversal`'s constructor as default, it would be nice to provide an `apply` alias for it.

core/shared/src/main/scala/monocle/Lens.scala
julien-truffaut: maybe something like `fromStore`?

PR #506, opened 2017-03-05
update CrossVersion.patch and scalaOrganization.value

aoiroaoino:
ref. https://github.com/typelevel/scala/issues/135


julien-truffaut:
👍 thanks @aoiroaoino

PR #508, opened 2017-03-16
support Scala Native

aoiroaoino:
first support: core, state
```
> show coreNative/nativeMissingDependencies
[info] * @java.net.URI
[info] * @java.net.URI::init_class.java.lang.String
[info] * @java.net.URI::toString_class.java.lang.String
[success] Total time: 4 s, completed 2017/03/17 0:27:58
> show stateNative/nativeMissingDependencies
[info] *
[success] Total time: 1 s, completed 2017/03/17 0:28:05
> show testNative/nativeMissingDependencies
[info] *
[success] Total time: 1 s, completed 2017/03/17 0:28:16
```
we need remove `java.net.URI` ...? :thinking:  /cc @densh @MasseGuillaume
https://github.com/julien-truffaut/Monocle/blob/v1.4.0/core/shared/src/main/scala/monocle/std/String.scala#L35-L36

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,3 +44,17 @@ install:
   - gem install jekyll -v 3.2.1
 after_success:
 - '[[ $TRAVIS_BRANCH == "master" && "${TRAVIS_PULL_REQUEST}" == "false" ]] && { sbt +publish ghpagesPushSite; };'
+matrix:
+  include:
+  - scala: 2.11.8
+    jdk: oraclejdk8
+    sudo: required
+    before_install:
+    - wget -O - http://llvm.org/apt/llvm-snapshot.gpg.key | sudo apt-key add -
+    - sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
+    - sudo add-apt-repository -y "deb http://apt.llvm.org/precise/ llvm-toolchain-precise main"
+    - sudo add-apt-repository -y "deb http://apt.llvm.org/precise/ llvm-toolchain-precise-3.7 main"
+    - sudo apt-get update -qq
+    - sudo apt-get install -y libgc-dev clang++-3.7 llvm-3.7 llvm-3.7-dev llvm-3.7-runtime llvm-3.7-tool libunwind7-dev
+    script:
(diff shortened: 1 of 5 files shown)


julien-truffaut:
Nice! About `java.net.URI` does it mean it would blow up at runtime or compile time if you try to use `URI` natively?

xuwei-k:
I think compile time

aoiroaoino:
runtime ...?
```
object Main {
  ...
  val uri = monocle.std.string.stringToURI.getOption("http://example.com")
}
```
```
> testNative/compile
[info] Compiling 1 Scala source to /Users/aoino/git/Monocle/testNative/target/scala-2.11/classes...
[success] Total time: 1 s, completed 2017/03/17 20:00:50
> testNative/run
[info] Linking NIR (623 ms)
[error] cannot link: @java.net.URI
[error] cannot link: @java.net.URI::init_class.java.lang.String
[error] cannot link: @java.net.URI::toString_class.java.lang.String
[error] unable to link
[error] (testNative/*:nativeLink) unable to link
[error] Total time: 1 s, completed 2017/03/17 20:00:53
```

xuwei-k:
The `run` task depends on the `nativeLink` task

- https://github.com/scala-native/scala-native/blob/v0.1.0/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePlugin.scala#L32
- https://github.com/scala-native/scala-native/blob/v0.1.0/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala#L406


Your `cannot link` error is during `nativeLink` task, not runtime.

- compile or package: create `.nir` files from "Scala sources". no errors if there is MissingDependencies
- nativeLink: create executable binary file from `.nir` files. error if there is MissingDependencies

julien-truffaut:
Ok so it means, you will not be able to use `stringToURI` in native but as you cannot deal with `java.net.URI` natively, I don't think it will be an issue.

julien-truffaut:
👍

aoiroaoino:
@julien-truffaut @xuwei-k 
thanks!

PR #513, opened 2017-04-01
Prefer `pure` over `point`.

sellout:
`pure` is also used in Cats, so it reduces the diffs between Scalaz and Cats branches.


julien-truffaut:
👍 thanks Greg

PR #515, opened 2017-04-04
remove scalaz data types in bench to simplify differences with cats branch

julien-truffaut:
@sellout it is your call if it is worth it or not


sellout:
This looks great to me. 👍🏾

PR #524, opened 2017-05-22
Allow for annotations on companion objects of @Lenses-annotated case classes.

### LINKED ISSUE #523: @Lenses only works if the case class has an uninteresting companion
Annotations or modifiers or whatnot on the companion break it.

```
scala> import monocle._, monocle.macros._
import monocle._
import monocle.macros._

scala> class annot extends annotation.StaticAnnotation
defined class annot

scala> @Lenses case class Foo(i: Int); @annot object Foo
<console>:24: error: Invalid annotation target: must be a case class
       @Lenses case class Foo(i: Int); @annot object Foo
        ^
```

hrhino:
Does what it says, I'd hope.

Fixes #523.


julien-truffaut:
Thanks @hrhino 👍

PR #528, opened 2017-08-12
Port to Cats.

sellout:
This is done in a style that attempts to minimize the diff from the Scalaz version, for as long as a Scalaz branch is supported.


julien-truffaut:
Thank you so much Greg, I will try to review it and check why tests are failing this weekend.

sellout:
The test that’s failing here is `Map.Each.get what you set`, which seems to be simply reversing the List that is the set value somehow. And it’s _only_ failing on the JVM (although locally it seems to be failing on only JS ¯\\\_(ツ)_/¯).

Also, Travis still seems to be running 2.10 and Scala Native jobs, even though I _thought_ I disabled those in both .travis.yml and build.sbt.

julien-truffaut:
I added the following unit test:
```scala
  test("each getAll") {
    val x = Map(1 -> "abc", 2 -> "cde")
    val e = each[Map[Int, String], String]
    e.modify(_.reverse)(x) should be (Map(1 -> "cba", 2 -> "edc"))
    e.getAll(x) should be (List("abc", "cde"))
    e.getAll(e.modify(_.reverse)(x)) should be (List("cba", "edc"))
  }
```

and I get:
```
[info] - each getAll *** FAILED ***
[info]   List("edc", "cba") was not equal to List("cba", "edc") (MapSpec.scala:18)
```

I suppose it comes from the Map `Traverse` instance in cats, it does not require the key to be ordered like scalaz does.

julien-truffaut:
even shorter example: 
```scala
test("each getAll") {
    val x = Map(1 -> "abc", 2 -> "cde")
    val y = Map(1 -> "cba", 2 -> "cba")
    val e = each[Map[Int, String], String]
    e.getAll(x) should be (List("abc", "cde"))
    e.getAll(y) should be (List("cba", "cba"))
}
```

```
[info] - each getAll *** FAILED ***
[info]   List("edc", "cba") was not equal to List("cba", "edc") (MapSpec.scala:21)
```

julien-truffaut:
I reported the issue in cats: https://github.com/typelevel/cats/issues/1831

julien-truffaut:
I changed the target branch to cats as I prefer to make the first few releases with cats dependency using a different artifact (e.g. `monocle-core-cats`).

Thank you very much Greg, it is really an amazing work.

PR #359, opened 2016-05-31
upgrade scalacheck and use CoGen to generate random functions

julien-truffaut:
see https://github.com/rickynils/scalacheck/issues/241

PR #529, opened 2017-08-16
Update to refined 0.8.2

fthomas:
This also makes use of refined-scalacheck instances in tests.


julien-truffaut:
Thanks @fthomas, I think we removed the scalacheck integration because we had some issues when we tried to upgrade scalacheck. Hopefully, scalacheck will be stable for a while.

fthomas:
I remember, it was removed during the transition from ScalaCheck 1.12 to 1.13 (https://github.com/julien-truffaut/Monocle/pull/359) but re-added in https://github.com/julien-truffaut/Monocle/pull/480.

PR #532, opened 2017-09-20
Add optics composition table

jdegoes:


julien-truffaut:
Thanks @jdegoes ! 

Good idea @kenbot about the re-ordering. I think John is busy for the next few days, I propose that we merge it as it is and re-order it in a separate PR. Someone on twitter also proposed to only fill half the matrix in order to show the symmetry

PR #533, opened 2017-09-27
Add state syntax for traversal

jeslg:
Adds syntax for `Traversal`s in the state module, including `extract`, `mod`, `assign` and related methods.

PR #534, opened 2017-10-01
use travis_setup.sh of scala-native

aoiroaoino:
https://github.com/julien-truffaut/Monocle/pull/533#issuecomment-333065488


julien-truffaut:
you're a genius!

PR #539, opened 2017-11-02
cats 1.0.0-RC1

tpolecat:
This updates Monocle to cats-1.0.0-RC1.

The only real change is `Map` → `SortedMap` everywhere, which is a hassle but is mostly mechanical. The other option would be to pull in alley-cats but personally I'd prefer to relegate those instances to a `monocle-alleycats` or something and keep the core lawful.

I deleted the aliased `IList` and `IMap` examples but left the rename imports like `{ Monoidal => Zip }`. If you'd like to get rid of these and other scalaz cruft I can do it here or in another PR.

### DIFF
--- a/test/shared/src/test/scala/monocle/std/MapSpec.scala
+++ b/test/shared/src/test/scala/monocle/std/MapSpec.scala
@@ -2,12 +2,14 @@ package monocle.std
 
 import monocle.MonocleSuite
 import monocle.law.discipline.function._
+import scala.collection.immutable.SortedMap
 
 class MapSpec extends MonocleSuite {
-  checkAll("at Map", AtTests[Map[Int, String], Int, Option[String]])
-  // typelevel/cats#1831 Map probably does not respect Traverse law at the moment
-  // checkAll("each Map", EachTests[Map[Int, String], String])
-  checkAll("empty Map", EmptyTests[Map[Int, String]])
-  checkAll("index Map", IndexTests[Map[Int, String], Int, String])
-  checkAll("filterIndex Map", FilterIndexTests[Map[Int, Char], Int, Char])
+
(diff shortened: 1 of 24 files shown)


tpolecat:
Updated the newts dep so we're green now.

tpolecat:
I checked with Travis re: circe optics and removal of `Map` instance and I think we might be ok.

> I think it might not be so bad since the `At`, `Each`, etc. for JSON objects uses our custom `traverse` on `JsonObject`
> …which respects order

test/shared/src/test/scala/monocle/std/MapSpec.scala
```diff
@@ -2,12 +2,14 @@ package monocle.std
...
-  checkAll("at Map", AtTests[Map[Int, String], Int, Option[String]])
-  // typelevel/cats#1831 Map probably does not respect Traverse law at the moment
-  // checkAll("each Map", EachTests[Map[Int, String], String])
-  checkAll("empty Map", EmptyTests[Map[Int, String]])
-  checkAll("index Map", IndexTests[Map[Int, String], Int, String])
-  checkAll("filterIndex Map", FilterIndexTests[Map[Int, Char], Int, Char])
+
+  checkAll("at SortedMap", AtTests[SortedMap[Int, String], Int, Option[String]])
+  // typelevel/cats#1831 SortedMap probably does not respect Traverse law at the moment
+  // checkAll("each SortedMap", EachTests[SortedMap[Int, String], String])
```
julien-truffaut: I guess we can uncomment it now

julien-truffaut:
Thanks a lot Rob!

tpolecat:
Cool. Can we get a milestone for this when it's convenient?

julien-truffaut:
I just released `1.5.0-cats-M2`

PR #544, opened 2017-11-06
replace if with filter option

sh0hei:


core/shared/src/main/scala/monocle/Getter.scala
```diff
@@ -18,7 +18,7 @@ abstract class Getter[S, A] extends Serializable { self =>
   /** find if the target satisfies the predicate  */
   @inline final def find(p: A => Boolean): S => Option[A] = s => {
     val a = get(s)
```
julien-truffaut: I think you could inline it

core/shared/src/main/scala/monocle/Iso.scala
```diff
@@ -63,7 +63,7 @@ abstract class PIso[S, T, A, B] extends Serializable { self =>
   /** find if the target satisfies the predicate  */
   @inline final def find(p: A => Boolean): S => Option[A] = s => {
     val a = get(s)
```
julien-truffaut: same here

julien-truffaut:
Thanks! 👍

PR #547, opened 2017-11-23
Map instance

julien-truffaut:


julien-truffaut:
@tpolecat if you have a min, would you mind to review this change

tpolecat:
This looks good to me. 👍

PR #550, opened 2017-12-20
Update cats to 1.0.0 and other dependencies

guersam:
Bumped scala, scala-js and scala-native versions as well.

PR #551, opened 2017-12-27
bump cats to 1.0.0

pepegar:
I'd benefit from this in [hammock](http://github.com/pepegar/hammock) :)


julien-truffaut:
Thanks @pepegar. However, we already have another PR to upgrade cats and newts: https://github.com/julien-truffaut/Monocle/pull/550

pepegar:
Awesome, thanks then. I was PR'ing newts too, but won't do it. 

cheers.

PR #555, opened 2018-01-12
update sbt to 1.0.4 (cats)

pepegar:
Update sbt to 1.0.4 and all the needed plugins.


julien-truffaut:
Generally, it is enough to make a PR to master and a maintainer will cherry pick it to cats branch. Thanks anyway!

pepegar:
Ah cool! I didn't know what process did you follow for updating `cats` from `master`.

julien-truffaut:
yeah it is normal, it is not documented :( 

We should probably update https://github.com/julien-truffaut/Monocle/blob/master/PROCESS.md

PR #560, opened 2018-03-22
helper methods to create instance of Each/Possible from more optics.

jbgi:


core/shared/src/main/scala/monocle/function/Each.scala
```diff
@@ -25,12 +25,28 @@ trait EachFunctions {
...
+  def apply[S, A](traversal: Traversal[S, A]) : Each[S, A] = new Each[S, A] {
+    override val each: Traversal[S, A] = traversal
   }
 
+  /** lift an instance of [[Each]] using an [[Iso]] */
+  def fromIso[S, A, B](iso: Iso[S, A])(implicit ev: Each[A, B]): Each[S, B] =
+    Each(iso composeTraversal ev.each)
+
+  /** lift an instance of [[monocle.Traversal]] using a [[Lens]] */
+  def fromLens[S, A, B](lens: Lens[S, A])(implicit ev: Each[A, B]): Each[S, B] =
```
julien-truffaut: what the intended usage of the other `fromX`? 

`fromIso` was mainly done for newtypes

julien-truffaut: you can but why not simply doing: `GenLens[RealUser](_.token) composeOptional possible`

jbgi: Yeah. Currently we have:
```scala
implicit val hasToken: Possible[RealUser, Token] = new Possible[RealUser, Token]  {
  val possible: Optional[RealUser, Token] = GenLens[RealUser](_.token) composePrism some
}
```
With the the `apply` helper method I could at least cut the boilerplate to:
```scala
  implicit val hasToken: Possible[RealUser, Token] = 
     Possible(GenLens[RealUser](_.token) composeOptional possible)
```
and yet a bit more with `fromLens`, but as you said I could live without it.

I can update this PR to only add the `apply` methods, if you prefer?

julien-truffaut: yes I would prefer that we only add the apply method

julien-truffaut: btw it might be good to do the same for all other typeclasses

jbgi: done!

core/shared/src/main/scala/monocle/function/Index.scala
```diff
@@ -27,6 +27,11 @@ trait IndexFunctions {
 }
 
 object Index extends IndexFunctions{
+
+  def apply[S, I, A](getOption: I => S => Option[A])(set: I => A => S => S): Index[S, I, A] = new Index[S, I, A] {
```
julien-truffaut: why did you inline the `Optional` definition? it makes the API inconsistent and potentially you can create Optional via other ways

jbgi: I did it like that to be consistent with `At.apply`. But I can change it to `I => Optional[S, A]` or maybe have both?

julien-truffaut: interesting, I completely forgot that we had `At.apply`. It is weird because it doesn't even look that we use it. 

Would you mind to add `I => Lens[S, A]` for `At` and use `I => Optional[S, A]` for `Index`? Maybe we can deprecate `At.apply` or add another apply to `Index` in another PR

jbgi: Done.
Apparently `At.apply` was used in `refined.BitsInstances`. I migrated the use to new `apply` method.

julien-truffaut:
👍

PR #573, opened 2018-05-30
sbt 1.1.6

xuwei-k:


julien-truffaut:
:+1:

julien-truffaut:
it seems that the build failed for 2.10
```
[error] (testJVM / Test / executeTests) error reading Scala signature of org.scalatest.package: value Predef is not a package
```

PR #577, opened 2018-06-29
update dependencies. drop Scala 2.10

xuwei-k:
- enable laws project compile with Scala 2.13.0-M4
- update discipline, refined, scalacheck and scalatest
- drop Scala 2.10. discipline 0.10.0 does not support Scala 2.10


julien-truffaut:
Thanks @xuwei-k :+1:

PR #578, opened 2018-07-02
Added `coProductIso`/`coProductDisjunctionIso` and `GenericOptics`

dcastro:
As discussed in the gitter channel, this PR adds an isomorphism between sum types (`shapeless.Coproduct`) and the disjunction of their parts.

If we have the following coproduct:

```scala
type ISB = Int :+: String :+: Boolean :+: CNil
```

We can use `coProductIso`/`coProductDisjunctionIso` to obtain the following isomorphisms:

```scala
val iso: Iso[ISB, Either[Int, Either[String, Boolean]]] = coProductIso[ISB].apply

val iso: Iso[ISB, Int \/ (String \/ Boolean)] = coProductDisjunctionIso[ISB].apply
```

## Use case

One use case where this is useful, is when every case class that extends a sealed trait has a common field, and you want to zoom in on it.

```scala
sealed trait X
case class A(name: String) extends X
case class B(name: String) extends X
case class C(name: String) extends X

val lens: Lens[X, String] =
  toGeneric[X] composeIso coProductDisjunctionIso.apply composeLens
    (GenLens[A](_.name) choice (GenLens[B](_.name) choice GenLens[C](_.name)))
```

## Notes

* `shapeless` gives us `CoproductToEither`/`EitherToCoproduct`. In order to be able to convert coproducts to scalaz's disjunction, I had to create two similar typeclasses in the `internal` package: `CoproductToDisjunction`/`DisjunctionToCoproduct`.
* I suggest that `monocle.generic.HListInstances.toHList` be deprecated, in favour of `monocle.generic.GenericIso.toGeneric` introduced in this PR.
* `CoproductToDisjunction` gives us a `\/` nested to the right, e.g. `A \/ (B (C \/ D))`. In the last code sample above, the user has to compose the lenses accordingly, e.g. `lensA choice (lensB choice (lensC choice lensD))`. The same thing happens if you use scalaz's `Choice.|||` operator. What are the owner's/maintainers' thoughts on adding a right-associative version of `|||` to make this composition a bit easier?

### DIFF
--- a/generic/shared/src/main/scala/monocle/generic/Generic.scala
+++ b/generic/shared/src/main/scala/monocle/generic/Generic.scala
@@ -1,0 +1,13 @@
+package monocle.generic
+
+import monocle.Iso
+import shapeless.Generic
+
+object generic extends GenericOptics
+
+trait GenericOptics {
+
+  /** An isomorphism between a type [[S]] and its generic representation. */
+  def toGeneric[S](implicit S: Generic[S]): Iso[S, S.Repr] =
+    Iso[S, S.Repr](S.to(_))(S.from(_))
+}
--- a/generic/shared/src/main/scala/monocle/generic/internal/CoproductToDisjunction.scala
+++ b/generic/shared/src/main/scala/monocle/generic/internal/CoproductToDisjunction.scala
@@ -1,0 +1,35 @@
+package monocle.generic.internal
+
+import scalaz.{\/, -\/, \/-}
+import shapeless._
+
+/**
+  * Typeclass converting a [[Coproduct]] to an [[\/]].
+  *
+  * Based on [[shapeless.ops.coproduct.CoproductToEither]]:
+  * https://github.com/milessabin/shapeless/blob/shapeless-2.3.3-scala-2.13.0-M4/core/src/main/scala/shapeless/ops/coproduct.scala#L1275-L1303
+  */
+sealed trait CoproductToDisjunction[C <: Coproduct] extends DepFn1[C] with Serializable
+
+object CoproductToDisjunction {
+  type Aux[In <: Coproduct, Out0] = CoproductToDisjunction[In] { type Out = Out0 }
+
+  implicit def baseToEither[L, R]: CoproductToDisjunction.Aux[L :+: R :+: CNil, L \/ R] = new CoproductToDisjunction[L :+: R :+: CNil] {
+    type Out = L \/ R
+    def apply(t: L :+: R :+: CNil): L \/ R = t match {
+      case Inl(l)         => -\/(l)
+      case Inr(Inl(r))    => \/-(r)
+      case Inr(Inr(cnil)) => cnil.impossible
+    }
+  }
+
+  implicit def cconsToEither[L, R <: Coproduct, Out0](implicit
+    evR: CoproductToDisjunction.Aux[R, Out0]
+  ): CoproductToDisjunction.Aux[L :+: R, L \/ Out0] = new CoproductToDisjunction[L :+: R] {
+    type Out = L \/ Out0
+    def apply(t: L :+: R): L \/ Out0 = t match {
+      case Inl(l) => -\/(l)
+      case Inr(r) => \/-(evR(r))
+    }
+  }
+}
--- a/example/src/test/scala/monocle/generic/CoproductExample.scala
+++ b/example/src/test/scala/monocle/generic/CoproductExample.scala
@@ -29,18 +29,33 @@ class CoproductExample extends MonocleSuite {
       illTyped("""coProductPrism[ISB, Float]""")
   }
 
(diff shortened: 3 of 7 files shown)


generic/shared/src/main/scala/monocle/generic/Generic.scala
```diff
@@ -0,0 +1,13 @@
+package monocle.generic
+
+import monocle.Iso
+import shapeless.Generic
+
+object generic extends GenericIso
+
+trait GenericIso {
```
dcastro: I wanted to name this trait `GenericInstances` to match the other traits, but unfortunately that name was already taken :/

generic/shared/src/main/scala/monocle/generic/internal/CoproductToDisjunction.scala
```diff
@@ -0,0 +1,35 @@
...
+
+object CoproductToDisjunction {
+  type Aux[In <: Coproduct, Out0] = CoproductToDisjunction[In] { type Out = Out0 }
+
+  implicit def baseToEither[L, R]: CoproductToDisjunction.Aux[L :+: R :+: CNil, L \/ R] = new CoproductToDisjunction[L :+: R :+: CNil] {
+    type Out = L \/ R
+    def apply(t: L :+: R :+: CNil): L \/ R = t match {
+      case Inl(l)      => -\/(l)
+      case Inr(Inl(r)) => \/-(r)
+      case _           => ???
```
julien-truffaut: what does it mean?

julien-truffaut:
Awesome work @dcastro ! 

Do you think it would be possible / desirable to fuse `toGeneric[X] composeIso coProductDisjunctionIso.apply`? Basically makes the `Generic.Rep` step hidden

generic/shared/src/main/scala/monocle/generic/internal/CoproductToDisjunction.scala
dcastro: This code was taken [straight from shapeless][0], but the missing pattern is `Inr(Inr(c))`, where `c` is of type `CNil`.

`CNil` is an unhinhabited type, it's only used to signal the end of a `:+:` list, so this pattern is guaranteed to never match.

However, I think replacing `case _ => ???` with `case Inr(Inr(cnil)) => cnil.impossible` would be more readable.

`CNil` has this `def impossible: Nothing` member that we're supposed to use it when pattern matching on it:

```scala
/** The CNil type is used to terminate a 'list' of :+: alternatives.
  *
  * Like the Nil constructor of List, it does not convey real information.
  * This is achieved by not having any value for CNil.
  *
  * This makes the type `Int :+: CNil` equivalent to `Int`, because the right (`Inr`) alternative
  * of `:+:` can not be constructed properly.
  */
sealed trait CNil extends Coproduct {
  /** Call this when you hit the CNil case in pattern matching to make the match exhaustive and safe. */
  def impossible: Nothing
}
```

I'll make this change.


 [0]: https://github.com/milessabin/shapeless/blob/shapeless-2.3.3-scala-2.13.0-M4/core/src/main/scala/shapeless/ops/coproduct.scala#L1275-L1303

dcastro:
Thanks 😄 !

I heartily agree that `toGeneric[X] composeIso coProductDisjunctionIso.apply` should be fused!
I'm not sure what to call that function though. Any suggestions?

Nevertheless, I think`toGeneric` and `coProductIso`/`coProductDisjunctionIso` should be public as is, there might be other use cases or other ways to compose them that I can't foresee.

On a related note, do you think we should have both `coProductIso` AND `coProductDisjunctionIso`? What's monocle's "policy" on `Either` vs `\/`?

julien-truffaut:
maybe we could have:

- `coProductToEither == toGeneric[X] composeIso coProductIso.apply`
- `coProductToDisjunction == toGeneric[X] composeIso coProductDisjunctionIso.apply`

In the master branch (using scalaz dependency), we generally have both `Either` and `Disjunction` but in the cats branch, we only support `Either`

dcastro:
Done. We should now have:

* `toGeneric`
* `coProductEitherIso`
* `coProductToEither`
* `coProductDisjunctionIso`
* `coProductToDisjunction`

generic/shared/src/main/scala/monocle/generic/Generic.scala
julien-truffaut: maybe we could call it `GenericOptics` in case we implement something else than `Iso` later but it doesn't really matter

julien-truffaut:
It looks great :+1: 

I will merge it over the weekend if no one is commenting. Thanks!

generic/shared/src/main/scala/monocle/generic/Generic.scala
dcastro: done, thanks!

PR #581, opened 2018-08-09
Added Override Modifiers

omervk:
It's a good and safe practice to add the override modifier to all overriding members.

Brought on by [wartremover-contrib's MissingOverride](https://github.com/wartremover/wartremover-contrib#missingoverride) triggering for the macro expansion of lenses.


julien-truffaut:
Thanks, could you also do the same for Prism and Iso macros?

julien-truffaut:
Thanks

omervk:
Thanks for merging this, @julien-truffaut. Any idea of when this might be officially released in a patch version?

julien-truffaut:
@omervk I am a bit busy atm but if someone can do the following it can be done fast:

1.    cherry pick commits from previous release
2.    update and run mima to check for binary compatibility
3.    create a release note

Thanks

PR #584, opened 2018-10-03
Documentation fixes

mattkohl:


julien-truffaut:
Thank you very much @mattkohl it is really useful

PR #609, opened 2018-11-14
Add optics for `cats.data.Chain` / `NonEmptyChain`

dcastro:
`Chain` doesn't have methods like `drop` and `updated` like lists do, so some of these implementations are a bit more verbose.

Also: deleted instances for cats' data types from `TestInstances`, and used the pre-defined instances from `import cats.laws.discipline.arbitrary._` instead.


julien-truffaut:
Awesome, thanks @dcastro 👍

PR #610, opened 2018-11-15
update Lens documentation about additional ways to compose

strobe:
I just would like to add a note about a few standard composition cases.

### DIFF
--- a/docs/src/main/tut/optics/lens.md
+++ b/docs/src/main/tut/optics/lens.md
@@ -102,6 +102,45 @@ val address = GenLens[Person](_.address)
 (address composeLens streetNumber).set(2)(john)
 ```
 
+## Other Ways of Lens Composition
+
+Is possible to compose few `Lenses` together by using `compose`:
+
+```tut:silent
+GenLens[Person](_.name).set("Mike") compose
+GenLens[Person](_.age).modify(_ + 1)
+```
+
+Same but with the simplified macro based syntax:
+
+```tut:silent
+import monocle.macros.syntax.lens.
+
+john
+  .lens(_.name).set("Mike") compose
+  .lens(_.age).modify(_ + 1)
+```
+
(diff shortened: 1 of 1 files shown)


docs/src/main/tut/optics/lens.md
```diff
@@ -101,6 +101,41 @@ val address = GenLens[Person](_.address)
 (address composeLens streetNumber).get(john)
 (address composeLens streetNumber).set(2)(john)
 ```
+### Other Ways of Lens Composition
+
+Is possible to compose few `Lenses` together by using `compose`:
```
julien-truffaut: it might be good to mention that all `Setter` like optics offer `set` and `modify` methods that returns an `EndoFunction` (i.e. `S` => `S`) which means that we can compose modification using basic function composition.

docs/src/main/tut/optics/lens.md
```diff
@@ -101,6 +101,41 @@ val address = GenLens[Person](_.address)
...
+`Sum` type - for that case you can compose `Prism` with `Lens` by using `some`:   
+
+```tut:silent
+import monocle.std.option.some
+
+case class A(b: Option[B])
+case class B(c: Int)
+
+val _b = GenLens[A](_.b)
+val _c = GenLens[Option[B]](_.map(_.c))
```
julien-truffaut: GenLens only work for case classes and fields of a case class, so it would not compile normally

docs/src/main/tut/optics/lens.md
```diff
@@ -101,6 +101,41 @@ val address = GenLens[Person](_.address)
...
+
+```tut:silent
+import monocle.macros.syntax.lens.
+
+john
+  .lens(_.name).set("Mike") compose
+  .lens(_.age).modify(_ + 1)
+```
+
+Sometimes you need an easy way to update `Product` type inside
```
julien-truffaut: Maybe worth mentioning that a more detailed view of the various optics composition is available here: http://julien-truffaut.github.io/Monocle/optics.html

julien-truffaut:
Sorry @strobe I somehow lost track of this PR. Thanks for the changes!

PR #621, opened 2019-01-09
Added `coProductIso`/`coProductDisjunctionIso` for cats version

maxkorolev:
This is a copy-paste pull request from #609 for cats version

I have an error during local publishing 
```
Monocle/generic/shared/src/main/scala/monocle/generic/GenericOptics.scala:10:3: Could not find any member to link for "S".
[error]   /** An isomorphism between a type [[S]] and its generic representation. */
```
Unfortunately, I didnt have enough time to resolve it, I see the same in master branch without my changes. Is there any proper way to solve it?

### DIFF
--- a/example/src/test/scala/monocle/generic/CoproductExample.scala
+++ b/example/src/test/scala/monocle/generic/CoproductExample.scala
@@ -1,5 +1,7 @@
 package monocle.generic
 
+import monocle.Lens
+import monocle.macros.GenLens
 import monocle.MonocleSuite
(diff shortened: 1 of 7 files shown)


julien-truffaut:
@MaxWorgan Thanks for the PR. However, we don't generally copy / paste feature between master and cats. The preferred method is to cherry pick commits and adapt it from scalaz to cats.

maxkorolev:
Ok, #623

PR #629, opened 2019-02-17
Fix implicit syntax collisions

vladimirkl:
This PR fixes a collision when we use something like [scala-newtype](https://github.com/estatico/scala-newtype) and have `import monocle.macros.syntax.lens._` in scope.  A cause of collision is `val value: A` in `GenApplyLensOps` so it conflicts with any `.value` that is defined via implicit syntax. The same issue was fixed in Cats code base - see https://github.com/typelevel/cats/issues/2491 and 
https://github.com/typelevel/cats/pull/2614.
Proposed change can be considered as binary compatible - unless someone used `GenApplyLensOps.value` in code which is unlikely. We can probably do a bit more and replace  `toGenApplyLensOps` and `GenApplyLensOps` with single implicit class - as Scala looks to be not instantiating it for macro based syntax. But this change can break some code using GenApplyLensOps directly, so for now I'm just making `.value` private and replace access to it in macro with pattern matching on tree.

### DIFF
--- a/core/shared/src/main/scala/monocle/Prism.scala
+++ b/core/shared/src/main/scala/monocle/Prism.scala
@@ -317,7 +317,7 @@ sealed abstract class PrismInstances {
   }
 }
 
-final case class PrismSyntax[S, A](self: Prism[S, A]) extends AnyVal {
+final case class PrismSyntax[S, A](private val self: Prism[S, A]) extends AnyVal {
 
   /** lift a [[Prism]] such as it only matches if all elements of `F[S]` are getOrModify */
   def below[F[_]](implicit F: Traverse[F]): Prism[F[S], F[A]] =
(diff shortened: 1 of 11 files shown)


julien-truffaut:
LGTM @vladimirkl thanks for the PR!

Do you know if it applies to any other macros? or other syntax classes?

julien-truffaut:
btw does this change only apply to cats branch? if not could you update the PR toward master, we will cherry pick it back to cats on the next realease.

vladimirkl:
I think that issue applies to every syntax class that subject as public field. Btw they are defined inconsistently - some are `AnyVal` and some are not. I can make them all `AnyVal` with private val subject. And PR can be renamed to "Fix implicit syntax collisions". I will try to make it against master.

julien-truffaut:
Thanks @vladimirkl !

vladimirkl:
@julien-truffaut should I make a PR for cats branch with cherry pick?

julien-truffaut:
it would be great if you have time

PR #645, opened 2019-04-17
enable Scala 2.13.0-RC1 build for all projects

xuwei-k:


julien-truffaut:
Nice, much simpler. Thank you @xuwei-k

PR #652, opened 2019-04-19
use SAM conversion

xuwei-k:


julien-truffaut:
good idea! Thanks Kenji

PR #653, opened 2019-04-20
disable the "release" stage if forked repo

xuwei-k:


julien-truffaut:
Thanks @xuwei-k, you are on fire!

Issue #649, opened 2019-04-18
Fix compiler warning for 2.13


julien-truffaut:
After upgrading to 2.13 I noticed a bunch of warnings (deprecation?).

spf3000:
Hey I started work on this.  PR should follow soon.  Will be my first OS contribution if successful!

spf3000:
Might be slightly more involved than I thought - I think we are going to need to put in separate folders for scala 2.12 and scala 2.13 because some of the e.g. https://blog.ssanj.net/posts/2017-06-26-crosscompile-multiple-versions-of-scala-via-sbt.html
I thought I had better check with the maintainers as I don't think that there are any so far unless I am mistaken?

julien-truffaut:
Thanks @spf3000 for looking into this issue. I just merged https://github.com/julien-truffaut/Monocle/pull/646 which makes scala 2.13 the "main" scala version for monocle. You might need to create a scala 2.12 folder to fix some warnings.

Maybe you can do it in two steps:
* fix all the warning you can without scala specific folder
* create a 2.12 specific folder for things like Stream who seem to have changed a lot.

spf3000:
sounds like a plan :)

julien-truffaut:
Hi @spf3000 , some of the warnings you may have been working on have been fixed in #651. Sorry about that it is rare that we get two people working on the same issue.

spf3000:
No problems @julien-truffaut. Some of the compiler warnings are deprecation warnings for Stream, which is deprecated in favour of LazyList. I could change to LazyList, only problem is you are using scalaz-Tree in (Monocle/core/shared/src/main/scala/monocle/std/Tree) which is implemented in terms of Stream - so it might be a little difficult to move off Stream for the moment - some of the other warnings are easy to fix though - I can definitely reduce the number of warnings.

julien-truffaut:
@spf3000 sounds good, let's tackle Stream separately

spf3000:
Hey I have a commit ready - but I get permission denied - 
`ERROR: Permission to julien-truffaut/Monocle.git denied to spf3000.
fatal: Could not read from remote repository.`
Do i need to be added as  a contributor or something?

julien-truffaut:
Hi @spf3000 ah yes that's because we use the fork - pull request model. I thought we had some links in the project but I can't find it anymore, you can have a look at https://gist.github.com/Chaser324/ce0505fbed06b947d962

[...]

PR #655, opened 2019-04-23
first attempt Issue #649 fix compiler warnings

spf3000:


core/shared/src/main/scala/monocle/function/Snoc.scala
```diff
@@ -57,12 +57,12 @@ object Snoc extends SnocFunctions {
       case (init, last) => init :+ last
     }
   )
-
-  implicit def streamSnoc[A]: Snoc[Stream[A], A] = Snoc(
-    Prism[Stream[A], (Stream[A], A)]( s =>
+// replace with LazyList
+  implicit def lazyListSnoc[A]: Snoc[LazyList[A], A] = Snoc(
```
julien-truffaut: I am not sure we can change this to LazyList yet because 2.12 doesn't have LazyList (if I am not mistaken)

core/shared/src/main/scala/monocle/std/Tree.scala
```diff
@@ -15,19 +14,19 @@ trait TreeOptics {
   final def rootLabel[A]: Lens[Tree[A], A] =
     Lens[Tree[A], A](_.rootLabel)(l => tree => Node(l, tree.subForest))
 
-  final def subForest[A]: Lens[Tree[A], Stream[Tree[A]]] =
-    Lens[Tree[A], Stream[Tree[A]]](_.subForest)(children => tree => Node(tree.rootLabel, children))
+  final def subForest[A]: Lens[Tree[A], LazyList[Tree[A]]] =
```
julien-truffaut: scalaz still seem to be using Stream, so I think we are stuck with it. I guess it is fine because we are dropping support for scalaz after the next release

julien-truffaut:
Hi @spf3000 , I added a couple of comments in line. Also, why did you move refined to a scala 2.13 folder?

spf3000:
Hi Julien,

So I totally stuffed up that push - something went wrong with the whole
commit/fork change origin thing and I committed some explorations I did
around Stream even though I decided to abandon them.  I will fix and
recommit.  Possibly just put it on a new branch. Sorry!

julien-truffaut:
> Hi Julien, So I totally stuffed up that push - something went wrong with the whole commit/fork change origin thing and I committed some explorations I did around Stream even though I decided to abandon them. I will fix and recommit. Possibly just put it on a new branch. Sorry!

no worry, I am almost always screw up and I have done it a lot ;)

PR #656, opened 2019-04-23
remove non-Stream compiler warnings

spf3000:


julien-truffaut:
@spf3000 nice but why did you move refined to a 2.13 folder?

spf3000:
Hi Julien,

Thanks, Yes obviously there is a bit of work left to do - I will have
another go o get the tests compiling.  I moved refined to a 2.13 folder
because of the changes to Bits.scala which won't compile in 2.13. (In order
to remove the compiler warnings to for signum (deprecated for sign).  So it
now won't compile under 2.12.  So I guess we can either - just make a copy
of Bits.scala under 2.12 and 2.13 - but the problem is it's a bit
fragmented from the rest of the code - or make wholesale moves to 2.13 -
which is what I have started to do by moving all of refined. I'm thinking
now the first option might be better, as if we go down this route we are
going to have 2 copies of the codebase that are mostly the same.

julien-truffaut:
Is `sign` available in 2.12? If yes, let's use `sign` otherwise I am not sure it is worth to split the code base just to avoid a warning. Maybe it is possible to create just a copy of `Bits.scala` for 2.12, I am not sure we need to duplicate all files depending on Bits

spf3000:
It's not available in 2.12. But yeah I will just create a copy of files
with a non-backwards-compatible change like Bits. It will be like

spf3000:
...cont
refined/scala-2.12/other-packages/Bits
refined/scala-2.13/other-packages/Bits
refined/scala/rest of refined

PR #658, opened 2019-04-24
Issue 649 d

spf3000:
So I realised that in order to implement fixes that aren't backwards compatible we are going to have to duplicate at least significant portions of the code base, if not all of it.  It's probably not a price you are willing to pay, understandably.  Hence only the ones (2 of them!) with a backwards compatible fix have been added.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,10 @@ target/
 project/project/
 benchData/
 _site/
+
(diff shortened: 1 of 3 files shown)


julien-truffaut:
Thanks @spf3000 for investigating it and nice quick win :)

PR #661, opened 2019-05-02
remove newts dependency, remove Wrapped typeclass

julien-truffaut:
I will stop maintaining newts. I hope there will be a better newtype story with opaque types

/cc @monadplus


julien-truffaut:
I officially stopped supporting newts

PR #662, opened 2019-05-04
kind-projector org change (and sbt version bump)

SethTisue:
it would helpful in the Scala community build to have this merged,
normally we can override org changes like this in our config, but I'm
having trouble I don't understand. and anyway, using the latest
versions is generally a good thing

the sbt bump isn't a necessary part of this, but it may help avert
unexpected community build trouble as well


julien-truffaut:
Thank you Seth.

PR #671, opened 2019-06-18
Better composePrism/composeIso for POptional

danslapman:
Motivation: `POptional` composed by current inplementation is unable to `set` a value properly if it points to `None`, because it calls `modify` underhood. `reverseGet` of PPrism always returns a valid value to set, so it's better not to convert PPrism to POptional for composition.

Please tell, if it is necessary to open a separate PR agains `cats` branch (I actually use cats version).

### DIFF
--- a/core/shared/src/main/scala/monocle/Optional.scala
+++ b/core/shared/src/main/scala/monocle/Optional.scala
@@ -162,7 +162,22 @@ abstract class POptional[S, T, A, B] extends Serializable { self =>
 
   /** compose a [[POptional]] with a [[PIso]] */
   @inline final def composeIso[C, D](other: PIso[A, B, C, D]): POptional[S, T, C, D] =
-    composeOptional(other.asOptional)
+    new POptional[S, T, C, D]{
+      def getOrModify(s: S): T \/ C =
+        self.getOrModify(s).map(other.get)
+
+      def set(d: D): S => T =
+        self.set(other.reverseGet(d))
+
+      def getOption(s: S): Option[C] =
+        self.getOption(s) map other.get
+
+      def modifyF[F[_]: Applicative](f: C => F[D])(s: S): F[T] =
+        self.modifyF(other.modifyF(f))(s)
+
+      def modify(f: C => D): S => T =
+        self.modify(other.modify(f))
+    }
 
   /********************************************/
   /** Experimental aliases of compose methods */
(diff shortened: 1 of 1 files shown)


julien-truffaut:
@danslapman Thanks, could you add an example/test to show the issue

danslapman:
@julien-truffaut sure

```scala
sealed trait AB
case object A extends AB
case object B extends AB

object AB {
  val a: Prism[AB, A] = ...
}

case class Bar(foo: Option[AB])
object Bar {
  val foo: Optional[Inner, AB] = ..
}

..

val barA = Bar.foo ^<-? AB.a

val bar = Bar(None)

barA.set(A)(bar) //Currently will do nothing
```

danslapman:
I will fix `composeIso` shortly as well.

julien-truffaut:
Thanks unfortunately this is the intended (lawful behavior). I am on my phone but I believe you will get an error if you check your Optional is lawful.

danslapman:
@julien-truffaut thank You, got it.

julien-truffaut:
@danslapman please let me know if you want a more detail explanation. Almost everyone gets bitten by this.

danslapman:
@julien-truffaut in fact it's pretty convenient to have a composasble `Optional`-like optic, that is able to set a value, `Lens[A, Option[B]]` does not compose. Maybe intoduce some lawless "OptionalInsert" in `monocle-unsafe`?

julien-truffaut:
IMO this law is one of the most fundamental, so I would rather not include this kind of lawless Optional.

danslapman:
Okay

danslapman:
@julien-truffaut then, maybe add some optic that can only `set`?

julien-truffaut:
it would not integrate with the existing optic stack. E.g. A `Setter[S, A]` cannot be transformed to a `SetOnly[S, A]` because it is `set` will have a different behaviour.

Optics are abstractions to zoom into 0, 1, n values of the same type. They follow the Functor / Traverse law that says that an Optic / Functor / Traverse cannot affect the structure, only the value. Here you are trying to update the structure because you want to "create" something. The only way to do that with optics is if you have an Optic[S, Option[A]] in this case you case `set(Some(a))`

danslapman:
>`Setter[S, A]` cannot be transformed to a `SetOnly[S, A]`
but it can be composed

I mean something like that:
```scala
abstract class PInjector[S, T, A, B] extends Serializable { self =>
  /** set polymorphically the target of a [[PInjector]] with a value */
  def set(b: B): S => T

  def ^|-:[W, G](other: PLens[W, G, S, T]): PInjector[W, G, A, B] =
    b => other.modify(set(b))
}
```

PR #676, opened 2019-06-24
Cats cherry picks

cquiroz:
This PR cherry picks most commits on master and applies them to the cats branch.
There is a significant amount of changes to the build so the approach taken was to ignore all commits related to 2.13.x and scala native and harmonize the builds at the end

I also updated to cats 1.6.1 however I had to comment out several tests. There is a bad interaction across the cats/discipline/scalacheck/scalatest version that make any tests that use `cats` arbitraries to fail with a binary incompatibility error

Scala 2.13 is not supported with cats 1.6.1, so only 2.12 is supported in this branch

I think the best approach is to soon update to cats 2.0.x and there support 2.13.0 and re enable the tests. However that maybe better done in another PR. I briefly tested 2.0.x that but there are some deprecation errors that need to be fixed to properly run on 2.13


julien-truffaut:
That's amazing, thank you very much @cquiroz

cquiroz:
And I'm thrilled to help on one of my favorite libs. I'd like to do the cats-2.0.0 migration as well

julien-truffaut:
@cquiroz I created a checklist for the cat migration: https://github.com/julien-truffaut/Monocle/issues/677

I diffed master and cats branch and it seems that nothing is missing. Great work!

PR #678, opened 2019-06-25
Upgrade to cats 2.0.0-M4

cquiroz:
This PR starts the ball on upgrading to cats 2.0.0, There are tasks pending:

- [x] re-enable commented tests on cats branch
- [x] remove type alias like import scala.{Left => -\/}
- [x] remove deprecated methods for scala 2.13


core/shared/src/main/scala/monocle/Fold.scala
```diff
@@ -68,22 +67,22 @@ abstract class Fold[S, A] extends Serializable { self =>
     !isEmpty(s)
 
   /** join two [[Fold]] with the same target */
-  @inline final def choice[S1](other: Fold[S1, A]): Fold[S \/ S1, A] =
-    new Fold[S \/ S1, A]{
-      def foldMap[M: Monoid](f: A => M)(s: S \/ S1): M =
+  @inline final def choice[S1](other: Fold[S1, A]): Fold[Either[S, S1], A] =
+    new Fold[Either[S, S1], A]{
```
cquiroz: The direct translation of `S \/ S1` is `S Either S1` but I think it would be easier to stick to `Either[S, S1]`

core/shared/src/main/scala/monocle/std/Validated.scala
```diff
@@ -0,0 +1,33 @@
+package monocle.std
+
+import monocle.{Prism, Iso, PIso, PPrism}
+
+import cats.data.Validated
+import cats.syntax.either._
+import cats.syntax.validated._
+
+object validation extends ValidatedOptics
```
cquiroz: I renamed these to `Validated` to follow cats naming, let me know if it is fine. This is a breaking change

generic/shared/src/main/scala/monocle/generic/internal/CoproductToDisjunction.scala
```diff
@@ -1,10 +1,9 @@
 package monocle.generic.internal
 
-import scala.{Either => \/}
 import shapeless._
 
 /**
-  * Typeclass converting a [[Coproduct]] to an [[\/]].
+  * Typeclass converting a [[Coproduct]] to an [[Either]].
```
cquiroz: I think this comment was originally incorrect

test/jvm/src/test/scala/monocle/PlatformSpecificTestInstances.scala
```diff
@@ -5,7 +5,9 @@ import org.scalacheck.{Arbitrary, Cogen, Gen}
 import cats.Eq
 
 private [monocle] trait PlatformSpecificTestInstances {
-  implicit val urlEqual = Eq.fromUniversalEquals[URL]
+  implicit val urlEqual: Eq[URL] = Eq.instance {
```
cquiroz: This `Eq` instance is bad as `equals` for `java.net.URL` is well known as blocking. It will try to do a dns lookup on the url making it very slow and it is also side effecting. The proper way should be to use URI instead

julien-truffaut: ah good point, I learn that recently from john degoes tweet. I didn't realise we were using it in Monocle

cquiroz:
A question is how to handle deprecated classes, in particular `Stream`. I sent a last commit with an alternative which is to have scala 2.12/2.13 alternatives for instances. In the 2.12 we can put the `Stream` instances, and we could later on add instances for `LazyList` on the 2.13 variant.

This is still incomplete as we need to do something similar for the tests but wanted to check your opinion about this approach. It is slightly complicated but preserves the instances for either version

julien-truffaut:
@cquiroz it is definitely the best approach for the end user but it is painful to put in place and maintain. You did the hard work of setting it up so, I will say let's go for it.

cquiroz:
I'm having troubles with these tests on 2.13.0
https://github.com/julien-truffaut/Monocle/blob/master/test/shared/src/test/scala/monocle/IsoSpec.scala#L130-L144
https://github.com/julien-truffaut/Monocle/blob/master/test/shared/src/test/scala/monocle/PrismSpec.scala#L156-L170

These tests rely on `Iso` or `Prism` equality which are not defined. These work on 2.12 but break on 2.13
Any opinions? I'd think to just remove them

cquiroz:
Also some examples don't compile on 2.13 due to the use of `Stream`. It is probably not worth supporting that on examples so I'lldelete them

julien-truffaut:
@cquiroz Sorry I am extremely busy until next Monday. I will try to review today. In the worst case, we don't have to fix everything in one PR. We can merge this one and fix the remaining issues next week.

Sorry for the inconvenience

cquiroz:
No problem whenever you can

build.sbt
```diff
@@ -37,47 +37,63 @@ inThisBuild(List(
...
+    case Some((2, y)) if y <= 12 =>
+      extraDirs("-2.12-")
+    case Some((2, y)) if y >= 13 =>
+      extraDirs("-2.13+")
+    case _ => Nil
+  }
+}
+
 lazy val buildSettings = Seq(
   scalaVersion       := "2.12.8",
```
julien-truffaut: shall we bump the main scala version to 2.13?

core/shared/src/main/scala/monocle/std/Validated.scala
julien-truffaut: should the object be called validated as well?

test/shared/src/test/scala-2.12-/monocle/std/CofreeSpec.scala
```diff
@@ -5,10 +5,10 @@ import monocle.MonocleSuite
 // import monocle.law.discipline.function._
 //
 // import cats.free.Cofree
-
+//
 class CofreeSpec extends MonocleSuite {
```
julien-truffaut: should we enable this test?

test/shared/src/test/scala/monocle/IsoSpec.scala
```diff
@@ -123,21 +123,20 @@ class IsoSpec extends MonocleSuite {
...
   }
 
-  test("GenIso nullary equality") {
-    GenIso.unit[Nullary] shouldEqual _nullary
-  }
-
-  test("GenIso unary equality") {
-    GenIso[Unary, Int] shouldEqual _unary
-  }
+  // test("GenIso nullary equality") {
```
julien-truffaut: Here are the eq instance: https://github.com/julien-truffaut/Monocle/blob/cats/test/shared/src/test/scala/monocle/TestInstances.scala#L50-L54

test/shared/src/test/scala/monocle/TestInstances.scala
```diff
@@ -25,38 +23,25 @@ trait TestInstances extends PlatformSpecificTestInstances with cats.instances.Al
...
+  implicit val quintaryEq      = Eq.fromUniversalEquals[Quintary]
 
-  implicit def optionCofreeEq[A](implicit A: Equal[A]): Equal[Cofree[Option, A]] =
-    Equal.instance { (a, b) =>  A.eqv(a.head, b.head) && a.tail === b.tail }
+  implicit def optionCofreeEq[A](implicit A: Eq[A]): Eq[Cofree[Option, A]] =
+    Eq.instance { (a, b) =>  A.eqv(a.head, b.head) && a.tail === b.tail }
 
-  implicit def streamCofreeEq[A](implicit A: Equal[A]): Equal[Cofree[Stream, A]] =
-    Equal.instance { (a, b) =>  A.eqv(a.head, b.head) && a.tail === b.tail }
+  implicit def pisoEq[S, T, A, B](implicit StoA: Eq[S => A], BtoT: Eq[B => T]): Eq[PIso[S, T, A, B]] =
```
julien-truffaut: I am not sure why the test with Iso/Prism equality do not compile, you seem to have kept these test instances

julien-truffaut:
@cquiroz looks great, I just made a couple of minor comments.

Do you think we should add a 2.13 folder with LazyList instead of Stream? If yes, probably in another PR

build.sbt
cquiroz: 👍

core/shared/src/main/scala/monocle/std/Validated.scala
cquiroz: I agree

cquiroz:
I'd add `LazyList` in a separate PR

julien-truffaut:
Amazing work @cquiroz thanks a lot

PR #682, opened 2019-07-02
Make some subprojects pure

cquiroz:
I noted that all cross projects are `Full` while they can be `Pure`. In fact only `core` and `test` need to be full.

In this PR I changed the types and moved the source files to the correct location

This change is not strictly needed and it is partly a matter of taste but it makes sense to me


julien-truffaut:
LGTM

julien-truffaut:
Thanks @cquiroz

PR #720, opened 2019-10-10
add Option and Either Prism in Prism's companion object

julien-truffaut:
do you think it makes sense to define them in the companion object of Prism?


cquiroz:
It seems fine, they are a bit hard to find now living on `std`.

julien-truffaut:
I agree but `Iso` seems to be a less clear cut. Should we put `Iso[List[A], Vector[A]]` in the companion object of `Iso`? what about the other way around? should we do all Seq combinations?

cquiroz:
I also wondered about adding these on each Optic, considering that it could be nice to have a very small kernel

julien-truffaut:
Yes, it is tricky to know where to draw the line. 

IMO, the most important is that the kernel will be very stable. For example, I am not sure if adding a method to a companion object break binary compatibility.

PR #724, opened 2019-10-17
Remove sbt-release

cquiroz:
I think sbt-release isn't needed and could be removed. This assumes we have a single common version which I think is true for monocle.
Note the `jsEnv` setting set in the build is now the default in scala.js and can be removed


julien-truffaut:
Nice, thank you.

PR #725, opened 2019-10-17
Proof of concept for new typeclasses

julien-truffaut:
- Use single type parameter and Aux pattern for typeclass definition
- move optics to companion object of Lens/Prism/...


cquiroz:
What is the benefit of this encoding over e.g
`class Cons[S, A]`?

cquiroz:
I can't fully comment on the implications as this feels advanced to me but the code looks good 👍

PR #726, opened 2019-10-20
At - Index typeclass

julien-truffaut:
- I used the same encoding as for `FieldX` and `Cons`.
- Introduced inheritance between `At` and `Index`
- I wonder if we should have an `AtK` version for higher kinded `S`, e.g. `Atk[List, Int]` instead of `At[List[A], Int, A]`

### DIFF
--- a/kernel/shared/src/main/scala/monocle/Optional.scala
+++ b/kernel/shared/src/main/scala/monocle/Optional.scala
@@ -1,6 +1,6 @@
 package monocle
 
-import monocle.function.Cons
+import monocle.function.{Cons, Index}
 
 trait Optional[A, B] { self =>
   def getOption(from: A): Option[B]
@@ -15,6 +15,7 @@ trait Optional[A, B] { self =>
   }
 
   def composeLens[C](other: Lens[B, C]): Optional[A, C] = compose(other)
+  def composePrism[C](other: Prism[B, C]): Optional[A, C] = compose(other)
 }
 
 object Optional {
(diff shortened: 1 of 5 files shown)


kernel/shared/src/main/scala/monocle/Optional.scala
```diff
@@ -15,6 +15,7 @@ trait Optional[A, B] { self =>
   }
 
   def composeLens[C](other: Lens[B, C]): Optional[A, C] = compose(other)
+  def composePrism[C](other: Prism[B, C]): Optional[A, C] = compose(other)
```
cquiroz: Are you planning to keep the symbolic operators?

kernel/shared/src/main/scala/monocle/function/Index.scala
```diff
@@ -0,0 +1,33 @@
+package monocle.function
+
+import monocle.Optional
+
+import scala.util.Try
+
+trait Index[S] {
```
cquiroz: So it is a value of type `A` indxed by  a type `I`? I guess that would be applicable to list like and map like instances 👍

kernel/shared/src/main/scala/monocle/Optional.scala
julien-truffaut: I don't know. Maybe not in kernel but we can always have them defined in an extra module. What do you think? In scala 3, we should only have one compose function.

kernel/shared/src/main/scala/monocle/function/Index.scala
julien-truffaut: indeed, it is even applicable to a String

PR #727, opened 2019-10-20
Possible typeclass

cquiroz:
This is really just an experiment with the new encoding.
Please do thorough check as I'm not totally familiar with this style


kernel/shared/src/main/scala/monocle/Optional.scala
```diff
@@ -28,4 +29,7 @@ object Optional {
 
   def tailOption[S](implicit ev: Cons[S]): Optional[S, S] =
     ev.tailOption
+
+  def possible[S, A](implicit ev: Possible.Aux[S, A]): Optional[S, A] =
```
cquiroz: Is this really useful?
A related question why is `headOption[S, A]: Optional[S, A]` while `tailOption[S]: Optional[S, S]`?

kernel/shared/src/main/scala/monocle/function/Possible.scala
```diff
@@ -0,0 +1,24 @@
...
+
+object Possible {
+  type Aux[S, A0] = Possible[S] { type A = A0 }
+
+  def apply[S, A0](optional: Optional[S, A0]): Aux[S, A0] = new Possible[S] {
+    type A = A0
+    override val possible: Optional[S, A0] = optional
+  }
+
+  implicit def optionPossible[A0](implicit ev: Optional[Option[A0], A0]): Possible[Option[A0]] =
```
cquiroz: I guess we don't want this `implicit` do we?

kernel/shared/src/main/scala/monocle/Optional.scala
julien-truffaut: I think it is. It is a way to summon a `possible` instance such as we can do:
```scala
foo compose Optional.possible
```

I believe we can offer a much better syntax in another module. I will send a proof of concept PR soon

kernel/shared/src/main/scala/monocle/function/Possible.scala
julien-truffaut: yeah here we should define an instance of the typeclass Possible for `Option[A]`. So the signature is most likely going to be:
```scala
implicit def option[A]: Possible.Aux[Option[A], A] = apply(...) 
```

kernel/shared/src/main/scala/monocle/Optional.scala
cquiroz: Do you think I should remove it?

julien-truffaut: ah no, I think it is useful

cquiroz:
Oh I did something bad with a rebase I'll redo the PR

PR #728, opened 2019-10-21
Proof of concept for new syntax module

julien-truffaut:
The idea is to propose a more concise way to compose optics for the most common cases, e.g. 

```scala
someLens._2.at(...).some.left
```

instead of 

```scala
someLens composeLens second composeLens at(...) composePrism some composePrism left
```

I also added `Applied`Lens/Prism/Iso/Optional that mirror the normal optic hierarchy but they are partially applied to an input, e.g.

```scala
foo.optic._1.index(...).set("hello")
```

instead of 

```scala
(first composeOptional index(...)).set("hello")(foo)
```

The applied syntax helps with type inference by providing the starting type of optic composition. In the above example, scalac cannot infer the type of `first`.


julien-truffaut:
Thanks, Carlos. I propose to remove 2.12 from the build at least for now. Anyway, this will not get released for a long time.

cquiroz:
Just noted this went to `master` rather than `3.x` is that fine?

julien-truffaut:
@cquiroz arf good catch

PR #732, opened 2019-10-24
Field3

cquiroz:
Mostly an experiment to get into 3.x
I had to disable mima due to https://github.com/lightbend/mima/issues/422


julien-truffaut:
LGTM, thanks Carlos.

PR #733, opened 2019-10-25
Harmonise type parameters

julien-truffaut:
I believe it is more idiomatic to use `A`, `B`, `C` or `A1`, `A2`, `A3` depending on the case. `S` comes from haskell lens `Lens S T A B` but we don't have to conform to it.


julien-truffaut:
I will redo a PR, I somehow screw up with the merge

PR #736, opened 2019-10-25
Fields

cquiroz:
Added Field4/Field5/Field6 matching Monocle 2.x
The issue with sbt-mima was solved upstream so I'm restoring it


julien-truffaut:
Great work. We are now on par with v2. Thanks @cquiroz

PR #738, opened 2019-10-26
Implement reverse

cquiroz:
Implementation of reverse

### DIFF
--- a/dotSyntax/src/main/scala/monocle/syntax/AppliedLens.scala
+++ b/dotSyntax/src/main/scala/monocle/syntax/AppliedLens.scala
@@ -1,7 +1,7 @@
 package monocle.syntax
 
 import monocle.Lens
-import monocle.function.{At, Field1, Field2, Field3, Field4, Field5, Field6}
+import monocle.function._
 
 trait AppliedLens[A, B] extends AppliedOptional[A, B]{
   def value: A
@@ -40,6 +40,9 @@ trait AppliedLens[A, B] extends AppliedOptional[A, B]{
 
   override def sixth(implicit ev: Field6[B]): AppliedLens[A, ev.B] =
     compose(ev.sixth)
+
+  override def reverse(implicit ev: Reverse[B]): AppliedLens[A, ev.B] =
+    compose(ev.reverse)
 
   override def asTarget[C](implicit ev: B =:= C): AppliedLens[A, C] =
     AppliedLens(value, optic.asTarget[C])
--- a/kernel/shared/src/main/scala/monocle/function/Reverse.scala
+++ b/kernel/shared/src/main/scala/monocle/function/Reverse.scala
@@ -1,0 +1,48 @@
+package monocle.function
+
+import monocle.Iso
+
+trait Reverse[A] {
+  type B
+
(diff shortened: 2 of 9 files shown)


kernel/shared/src/main/scala/monocle/function/Reverse.scala
```diff
@@ -0,0 +1,48 @@
...
+import monocle.Iso
+
+trait Reverse[A] {
+  type B
+
+  def reverse: Iso[A, B]
+}
+
+object Reverse {
+  type Aux[A, B0] = Reverse[A] { type B = B0 }
```
julien-truffaut: maybe worth to have a second type alias for the case where `A = B`

dotSyntax/src/main/scala/monocle/syntax/AppliedLens.scala
```diff
@@ -41,6 +41,9 @@ trait AppliedLens[A, B] extends AppliedOptional[A, B]{
   override def sixth(implicit ev: Field6[B]): AppliedLens[A, ev.B] =
     compose(ev.sixth)
 
+  override def reverse(implicit ev: Reverse[B]): AppliedLens[A, ev.B] =
```
julien-truffaut: I think you also need to override it for `AppliedPrism`

kernel/shared/src/main/scala/monocle/function/Reverse.scala
```diff
@@ -0,0 +1,48 @@
+package monocle.function
+
+import monocle.Iso
+
+trait Reverse[A] {
+  type B
+
+  def reverse: Iso[A, B]
```
julien-truffaut: can you also add it to the companion object of Iso

kernel/shared/src/main/scala/monocle/function/Reverse.scala
cquiroz: do you mean
`type Aux[A] = Reverse[A] {type B = A}`?
I guess it would need another name

julien-truffaut: yes, unfortunately, I don't have a good suggestion for names.

dotSyntax/src/main/scala/monocle/syntax/AppliedLens.scala
cquiroz: If I override it on `AppliedPrism` as
`  override def reverse(implicit ev: Reverse[B]): AppliedPrism[A, ev.B] =`
I start getting issues in `AppliedIso` as its parents `AppliedPrism` and `AppliedLens` both override `reverse` with different signatures. 
Any ideas?

julien-truffaut: got it, it is because of `reverse` in `AppliedIso` returns an `AppliedLens` while it should be an `AppliedIso`.

cquiroz: Thanks, I fixed that in the latest PR

julien-truffaut:
Great work.

PR #740, opened 2019-10-30
#739 Missing instance of At for ListMap

### LINKED ISSUE #739: Missing instance of At for ListMap
Hi, I'm playing around with monocle for adjusting a nested case class structure (generated openapi description from the tapir library) but I'm running into somewhat confusing compiler errors.

```scala
val d: OpenAPI = ???
val paths: Lens[OpenAPI, ListMap[String, PathItem]] = GenLens[OpenAPI](_.paths)
val parameters: Lens[PathItem, List[OpenAPI.ReferenceOr[Parameter]]] = GenLens[PathItem](_.parameters)
val parameterSchema: Lens[Parameter, OpenAPI.ReferenceOr[Schema]] = GenLens[Parameter](_.schema)

val x  = (paths composeLens at("/product/{id}")).getOption(d)
```

This will result in an ambiguous error.

```
[error] ...ambiguous implicit values:
[error]  both method atMap in object At of type [K, V]=> monocle.function.At[Map[K,V],K,Option[V]]
[error]  and method atSet in object At of type [A]=> monocle.function.At[Set[A],A,Boolean]
[error]  match expected type monocle.function.At[S,String,A]
[error]     val x  = (paths composeLens at("/product/{id}")).getOption(d)
```

I read the FAQ about it but was under the impression that the generated lens would be enough to avoid it, or am I wrong?

So far this seems to be caused by a missing instance of `At` for `ListMap`.

Regards,

Jens

jan0sch:
### Changes

* add `At` instance for `ListMap`
* fixes #739

### DIFF
--- a/core/shared/src/main/scala/monocle/function/At.scala
+++ b/core/shared/src/main/scala/monocle/function/At.scala
@@ -3,7 +3,7 @@ package monocle.function
 import monocle.{Iso, Lens}
 
 import scala.annotation.implicitNotFound
-import scala.collection.immutable.SortedMap
+import scala.collection.immutable.{ListMap, SortedMap}
 
 /**
  * Typeclass that defines a [[Lens]] from an `S` to an `A` at an index `I`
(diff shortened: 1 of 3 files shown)


julien-truffaut:
LGTM, thanks Jens

jan0sch:
Cool, I'll take a look.

julien-truffaut:
Great. Thanks Jens

PR #741, opened 2019-10-31
Possible implementation

cquiroz:
Take 2 on the implementation of `Possible`

### DIFF
--- a/kernel/shared/src/main/scala/monocle/function/Possible.scala
+++ b/kernel/shared/src/main/scala/monocle/function/Possible.scala
@@ -1,6 +1,6 @@
 package monocle.function
 
-import monocle.Optional
+import monocle.{Optional, Prism}
 
 trait Possible[A] {
   type B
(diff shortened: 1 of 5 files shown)


julien-truffaut:
Looks great, could you add a test in dot syntax?

kernel/shared/src/main/scala/monocle/function/Possible.scala
```diff
@@ -0,0 +1,22 @@
...
+
+object Possible {
+  type Aux[A, B0] = Possible[A] { type B = B0 }
+
+  def apply[A, B0](optional: Optional[A, B0]): Aux[A, B0] = new Possible[A] {
+    type B = B0
+    override val possible: Optional[A, B0] = optional
+  }
+
+  implicit def optionPossible[A](implicit ev: Optional[Option[A], A]): Aux[Option[A], A] =
```
julien-truffaut: As a general rule, optics should not be passed implicitly because they are not guaranteed to be unique per type. 

Here, I think we could do:
```scala
implicit def optionPossible[A]: Aux[Option[A], A] = apply(Prism.some)
```

cquiroz:
I added a test though I'm a bit unsure now. I'd appreciate a review

julien-truffaut:
Looks great Carlos, Thanks.

PR #742, opened 2019-11-02
Missing instance of Index for ListMap

jan0sch:
This adds a missing instance of `Index` for `ListMap` and should fix compilation errors like the one below.

```
[error] ...: ambiguous implicit values:
[error]  both method mapIndex in object Index of type [K, V]=> monocle.function.Index[Map[K,V],K,V]
[error]  and method sortedMapIndex in object Index of type [K, V]=> monocle.function.Index[scala.collection.immutable.SortedMap[K,V],K,V]
[error]  match expected type monocle.function.Index[S,String,A]
```


julien-truffaut:
Thanks @jan0sch

PR #743, opened 2019-11-05
Remove from build all unused modules and dependencies for v3 

julien-truffaut:
The goal is to experiment faster with syntax and kernel.

Should we also remove the source code of those modules?


cquiroz:
I check the old code from now and then but I guess we can do that switching branches

Issue #735, opened 2019-10-25
add field syntax to optics and applied optics


julien-truffaut:
related to #714 and #730 

Add `field` to syntax (`AppliedX` and `XOps` in dot-syntax module) such as we can generate Lens ala `GenLens`, e.g.

```scala
x.optic.field(_.foo.bar).some.field(_.buzz).set("foo")
```

Issue #714, opened 2019-09-29
Monocle 3


julien-truffaut:
Here is a couple of ideas I have for the next major version of Monocle:
1. Have a small and very stable core (e.g. optics + laws + typeclasses + instances, no syntax). It would allow other libraries to safely depend on Monocle instead of introducing their own optics (e.g. doobie).
1. Offer more friendly syntax for common cases, e.g. `lens[Foo](_.bar.fizz).some.at("hello")`. Dot syntax makes it discoverable by IDE, and it is concise.
1. Move away from personal repo to an organisation. I want to find other active maintainers and give them admin rights.
1. Use inheritance between optics to simplify conversions (e.g. Lens extends Optional)
1. Remove polymorphic optics

We could also go further with 1) and try to have a 0-dependency core. It will probably be a problem for Traversal because we use Van Laarhoven encoding. 

Please let me know what you think and if you have other ideas. Also, let me know if you would like to become a maintainer.

cquiroz:
A 0 dependency core would be a great objective. I think for example scalajs-react could benefit form that

I'd like to help as maintainer though I haven't contributed that much in the past. I'm a heavy Monocle user though

mmynsted:
I like idea 1. If you can reach a zero dependency core, so much the better.

neko-kai:
re: zero dependency core, Optional instances can be used to provide typeclass instances in separate modules without burdening users with wildcard importsh, see ttps://blog.7mind.io/no-more-orphans.html

julien-truffaut:
Thanks @neko-kai that might be useful to provide typeclass instance for cats or other data structure.

For reference, the main issue I see with the 0 dependency core is around `Traversal` encoding. We use the Van Laarhoven encoding where all `Traversal` methods are implemented in term of `modifyF`:

```scala
def modifyF[F[_]: Applicative](f: A => F[B])(s: S): F[T]
```

jdegoes:
Please zero dependency! 🙏 

And I'll volunteer to help on the architecture if my help is desired... and if not just say so, no need for the ban hammer! 😆 🙏

julien-truffaut:
All help is welcome, John :)

jdegoes:
Should I contribute here or somewhere else?

I have sketches of the following:

 - subtyping across optics
 - variance within optics
 - optic combinators
 - zero-dependency formulation of `modifyF`, among others

which may or may not be useful.

Excited about this!

[...]

PR #745, opened 2019-11-05
Field syntax for AppliedLens

julien-truffaut:
Partially implement #735

```scala
case class Foo(i: Int, bar: Bar)
case class Bar(b: Boolean, s: String)
val foo = Foo(5, Bar(true, "Hello"))

foo.optic.field(_.bar).field(_.b).get shouldEqual foo.bar.b
```

We should also be able to chain fields like `_.bar.b` but there is currently a compiler error.

We can repeat the process for all optics, applied optics.


cquiroz:
Maybe I can try to do the same for another of the optics, Prism perhaps?

julien-truffaut:
> Maybe I can try to do the same for another of the optics, Prism perhaps?

Sure. Do you mean add `field` to `Prism` / `AppliedPrism` or add another method on `AppliedLens` that simplifies composing with the branch of a sealed trait?

cb372:
I know a bit about macros and this looks reasonable to me 😄

julien-truffaut:
Thanks @cb372 for the review. Do you have an idea of how we can generalise that macro to other optics/applied optics without completely duplicating the code?

cb372:
Yes that should be do-able, but I'll have to have a look at the rest of the code. I haven't been following the progress of 3.x so I need to catch up. Is there any kind of design doc I should look at before I dive into the code?

julien-truffaut:
It is not really formalised. We have a couple of ideas but nothing really set in stone. This is probably the best plan we have so far: https://github.com/julien-truffaut/Monocle/issues/714.

Regarding this particular point, we are trying to have a convenient syntax to compose optics for the most common scenarios such as a user can do:

```scala
someOptioc.field(_.foo)
```

instead of this syntax in Monocle 2:

```scala
someOptic composeLens GenLens[Bar](_.foo)
```

We are trying to implement `field` this method on all optics (`Lens`, `Prism`, `Iso`, ...) and all applied optics (`AppliedLens`, `AppliedPrism`, `AppliedIso`).

I am not sure where these methods should be defined. Should we have them on optics/applied optics classes or should we "pimp" them on a separate module.

PR #747, opened 2019-11-08
Missing instance of Each for ListMap

jan0sch:
* add instance for `Each` using custom `Traversal` for `ListMap`

### DIFF
--- a/core/shared/src/main/scala/monocle/function/Each.scala
+++ b/core/shared/src/main/scala/monocle/function/Each.scala
@@ -3,6 +3,7 @@ package monocle.function
 import monocle.{Iso, PTraversal, Traversal}
 
 import scala.annotation.implicitNotFound
+import scala.collection.immutable.ListMap
 import cats.{Applicative, Order, Traverse}
 
 /**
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/monocle/function/Each.scala
```diff
@@ -54,6 +55,16 @@ object Each extends EachFunctions with EachInstancesScalaVersionSpecific {
 
   implicit def listEach[A]: Each[List[A], A] = fromTraverse
 
+  implicit def listMapTraversal[K, V]: Traversal[ListMap[K, V], V] = new Traversal[ListMap[K, V], V] {
```
julien-truffaut: could you inline this Traversal in the Each instance below? We only mark as implicit instances of typeclass but an optic is not a typeclass. For example, you could create a `Traversal[ListMap[K, V], V]` that only map the first and last element of the ListMap.

jan0sch: Done. :-)

julien-truffaut: cheers

PR #749, opened 2019-11-10
Fix chained field selection in macro

cb372:
Small fix to the GenLens macro to support `_.foo.bar`.

The generated code now calls `Lens#compose` (which returns a `Lens`) instead of `Optional#composeLens` (which returns an `Optional`).

### DIFF
--- a/macro/src/main/scala/monocle/macros/internal/Macro.scala
+++ b/macro/src/main/scala/monocle/macros/internal/Macro.scala
@@ -44,8 +44,8 @@ private[macros] class MacroImpl(val c: blackbox.Context) {
         )
       ) if termDefName.decodedName.toString == termUseName.decodedName.toString =>
         c.Expr[Lens[A, B]](
-          typesFields.map{ case (t,f) => q"monocle.macros.GenLens[$t](_.$f)" }
-                     .reduce((a,b) => q"$a composeLens $b")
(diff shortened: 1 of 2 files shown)


macro/src/main/scala/monocle/macros/internal/Macro.scala
```diff
@@ -44,8 +44,8 @@ private[macros] class MacroImpl(val c: blackbox.Context) {
         )
       ) if termDefName.decodedName.toString == termUseName.decodedName.toString =>
         c.Expr[Lens[A, B]](
-          typesFields.map{ case (t,f) => q"monocle.macros.GenLens[$t](_.$f)" }
-                     .reduce((a,b) => q"$a composeLens $b")
+          typesFields.map{ case (t,f) => q"_root_.monocle.macros.GenLens[$t](_.$f)" }
```
cb372: Including `_root_` in fully-qualified paths is a macro hygiene thing. It ensures that the generated code can't clash with anything else called `monocle` that happens to be in scope at the call site.

julien-truffaut: Nice, I didn't know why we had this in the code base.

julien-truffaut:
Thanks Chris!

PR #750, opened 2019-11-11
More syntax macros

cb372:
@julien-truffaut Is this the kind of thing you were thinking? I haven't tried to refactor the macro yet, so the code duplication is still there.


macro/src/test/scala/monocle/LensGenSpec.scala
```diff
@@ -1,26 +1,58 @@
 package monocle
 
 import monocle.macros.GenLens
-import monocle.macros.syntax.lens._
 import monocle.syntax.all._
+import monocle.macros.syntax._
```
julien-truffaut: I don't what is the "standard" import nowadays:

1. `.syntax._`
1. `.syntax.all._`
1. `.implicits._`

kernel/shared/src/main/scala/monocle/function/Cons.scala
```diff
@@ -20,13 +20,13 @@ object Cons {
       def cons: Prism[A, (B0, A)] = prism
     }
 
-  implicit def list[A]: Cons[List[A]] =
+  implicit def list[A]: Cons.Aux[List[A], A] =
```
julien-truffaut: good catch

julien-truffaut:
Yes, it looks great. I wonder if there is a way to reduce the macro duplication but it is not a big deal as long as the end-users get a nice API.

macro/src/test/scala/monocle/LensGenSpec.scala
cb372: In Cats at least, 3. means something different, as that brings in both syntax and instances.

If you want to allow users to opt-in to syntax at a very granular level, e.g. `syntax.lens._`, `syntax.prism._`, then `syntax.all._` probably makes more sense. But there are a couple of problems with this:

1. I think it's probably too granular. I can't think of a good reason why a user would want to enable the `field` method for e.g. lenses but not for other optics. So 99% of users will just use `syntax.all._` anyway

2. It didn't play nicely with the optics inheritance hierarchy (Iso extends both Lens and Prism, which both extend Optional) when I tried it. See my commit message for 0b70eb088ebf066a3a32e484da2ed16a3fcdff56. There's probably a way to structure the code to allow fine-grained syntax imports without introducing ambiguous implicits, but I couldn't think of a nice solution.

julien-truffaut: ah I didn't know about 3). I don't think we should be aiming for fine-grained imports, in practice, I almost never have seen anyone interested in that feature.

Btw, do you know if it is possible to have the field macro called straightaway from the optics/applied optics instead of using an implicit class? This way user would not need any additional import. Can we have macros in the same module as the optics module?

macro/src/main/scala/monocle/macros/syntax/FieldSyntax.scala
```diff
@@ -0,0 +1,59 @@
...
+
+  implicit class GenFieldsAppliedLens[A, B](private val value: AppliedLens[A, B]) {
+    def field[C]( f: B => C ): AppliedLens[A, C] = macro FieldSyntaxImpl.field_impl[AppliedLens, A, B, C]
+  }
+
+}
+
+class FieldSyntaxImpl(val c: blackbox.Context){
+
+  def field_impl[R[_, _], A, B: c.WeakTypeTag, C](f: c.Expr[B => C]): c.Expr[R[A, C]] = {
```
julien-truffaut: wow amazing

PR #751, opened 2019-11-13
Add PTraversal.parModifyF

nigredo-tori:
This relates to `PTraversal.modifyF` like `Parallel.parTraverse` relates to `Traverse.traverse`.

It's easy to define `parModifyF` for `PLenses`, `PPrism`s and `POptional`s, but I don't see the need for it yet.


julien-truffaut:
Nice, it looks like a great idea. I agree, it only seem useful for `Traversal`

Issue #417, opened 2016-11-12
Overloaded compose methods


julien-truffaut:
We cannot use a single compose method overloaded for each optics because type inference would be much weaker. Here is a scala ticket to summarize the issue: https://issues.scala-lang.org/browse/SI-10046

kenbot:
A possible alternative, without overloading: 

Optic composition is in effect a type function `O × P → Q` where `Q` is uniquely determined by the optic types `O` and `P`.  This might be modelled with type families and fundeps in Haskell.  Happily, as per [this article](https://wiki.haskell.org/GHC/Type_families#Type_class_instances_of_family_instances), we should be able to fake it with typeclasses and existential types.

Off the top of my head it seems that the following typeclass could therefore enable uniform composition:
```
sealed trait OpticCompose[O,P] {
  type Result
  def compose(o: O, p: P): Result
}
```
There would need to be a maximum of 8x8 = 64 instances for the different Monocle optics. This might do all kinds of violence to compile times, but who knows.  While the woeful polynomial explosion of composition relations remains, it is at least somewhat hidden from the poor user.

I'll try and knock up a proof of concept in a fork over the next day or so.

kenbot:
`@implicitNotFound("Sensible and informative error message")`

julien-truffaut:
@kenbot we tried a similar approach in https://github.com/julien-truffaut/OpticsExp/blob/master/src/main/scala/optic/lub/Lub.scala but we got the same issue with type inference 😞

kenbot:
On a hunch that the subtyping and F-bounded stuff might have had an impact I gave it a spin, but failed in the same way.  Ah well.

For the record:
```
sealed trait OpticCompose[Opt1[_,_,_,_], Opt2[_,_,_,_]] {
  type Result[_,_,_,_]

  def compose[S,T,A,B,C,D](o1: Opt1[S,T,A,B], o2: Opt2[A,B,C,D]): Result[S,T,C,D]
}
```
```
// Lens.scala
final def compose[Opt[_,_,_,_],C,D](opt: Opt[A,B,C,D])(implicit cmp: OpticCompose[PLens, Opt]): cmp.Result[S,T,C,D] = cmp.compose(this, opt)
```

Resulting in:
```
// SUCCEEDED:
basketItems composeTraversal each
res6: monocle.PTraversal[ShoppingBasket,ShoppingBasket,Item,Item] = monocle.PTraversal$$anon$2@75579ef7

[...]

PR #752, opened 2019-11-17
WIP: Sketch of possible design for 3.x

jdegoes:
Together with @adamfraser, we created a sketch of a possible design for Monocle 3.x.

The design:

 - **Reflects the optic hierarchy using subtyping**. Now an IDE can tell users that a lens is an optional, for example, without users having to remember or use conversion functions.
 - **Fully embraces declaration-site variance**. To use declaration-site variance, the fully polymorphic optic types are required; this, in general, will _improve_ type inference, despite the additional number of parameters.
 - **Simplifies implementations**. By making many methods `final`, duplication across optic types can be eliminated with default implementations that are comparable (allocation and dispatch-wise) to the specialized alternatives previously present.
 - **Removes all dependencies**. This pull request uses a simplified encoding for `Traversal`, which should encapsulate 99% of all use cases. It's worth debating whether or not to introduce a `Combine` interface (without implicits) that features `zipWith` and `make` methods to zip two structures together and make one from a value (which is wholly equivalent in terms of expressive power to the current encoding and can enable separate `interop` libraries to create these optics for any traversable).

We are not done but this is good enough for some initial feedback. Remaining work:

 - Perfect type inference so users can compose all optics using the same name, e.g.:
    ```scala
    _Org >>> _Site >>> _Manager >>> _Employees >>> _Salary
    ```
    Not having to choose a method name based on the target composition type will dramatically improve usability of the API and appearance of usage, and we have done a (somewhat unsatisfying) POC to show it's possible, but we look for better ways of doing it.
 - Currently using `Lens.first` and so forth without overloads results in type inference issues. Adam and I discussed how to solve this problem using a technique introduced in ZIO Test. It's quite possible this would result in flawless type inference in all cases at some cost for library maintainers. We have to POC it to see if it works and if the overhead is worth it.

Finally, there are a huge number of combinators that can be defined on these optics, which are not currently defined in Monocle 2.x, which can make the library much more powerful and user-friendly out of the box. However, before defining them or at least sketching them out, I think it's worth solidifying the core design.


julien-truffaut:
Thank you, John and Adam, for the pull request. I will need some times to go through it in details. Please find below my initial comments:

It is great you are trying to tackle the issue of non-overloaded compose methods (i.e. removing `composeLens`, `composePrism`). We tried many approaches over the year unsuccessfully. Your attempt looks quite similar to a spike I did with Edward Kmett a couple of years ago. The bottom line is that it is a bug/missing feature in scala 2 (https://github.com/julien-truffaut/Monocle/issues/417), and I verified it is fixed in dotty (https://github.com/julien-truffaut/Monocly).

You make a good case to keep the polymorphic optics. I don't think polymorphic optics are practical in Scala, but at least they would allow variance for Getter and Fold. I am not sure variance will improve type inference in the monomorphic case since type parameters are invariant in that scenario.

I also had a version with sealed hierarchy, but I ended up not going for it. In my opinion, an optic is an interface, not an enumeration. If we sealed the hierarchy, it would allow people to pattern match on optics which would break encapsulation. For example, if a function takes a Getter as an input, one could check if the Getter is a Lens or Iso and in that case, they could use set/reverseGet to modify. I am curious to know your arguments for sealing the hierarchy. 

Great implementation for Fold. I need to take more time to review Traversal. Could you elaborate on the limitations of your encoding (you seem to imply it has some)? For example, can you use a Traversal with Validated (to accumulate errors) or with IO?

I like `>>>` as an operator alias for `compose`. It looks like `>>>` is an alias for `andThen` in `Category`  maybe we should also rename optics `compose` function to `andThen` to be consistent.

Cheers

kernel/shared/src/main/scala/monocle/Iso.scala
```diff
@@ -2,37 +2,13 @@ package monocle
...
-  def apply[A, B](_get: A => B)(_reverseGet: B => A): Iso[A, B] =
-    new Iso[A, B] {
-      def get(from: A): B      = _get(from)
-      def reverseGet(to: B): A = _reverseGet(to)
-    }
+  def apply[S, A](_get: S => A)(_reverseGet: A => S): Iso[S, A] =
+    PIso[S, S, A, A](_get)(_reverseGet)
 
-  def reverse[A, B](implicit ev: Reverse.Aux[A, B]): Iso[A, B] =
+  def reverse[S, A](implicit ev: Reverse.Aux[S, A]): Iso[S, A] =
```
cquiroz: I think we standardized on using `A, B, ..` for type params rather than `S, A`

julien-truffaut: We might need to revisit that if we do go for polymorphic optics. We could have polymorphic optics with A1, A2, B1, B2, e.g. `PLens[A1, A2, B1, B2]` and `Lens[A, B]`

kernel/shared/src/main/scala/monocle/optics.scala
```diff
@@ -0,0 +1,197 @@
+package monocle
+
+sealed trait Fold[-S, +A] extends Serializable {
+  def fold[Z](zero: Z, combine: (Z, Z) => Z)(f: A => Z)(s: S): Z
+}
+
+sealed trait PFold[+LowerS, -UpperS, +LowerA, -UpperA] extends Serializable {
```
julien-truffaut: Do we need PFold?

kernel/shared/src/main/scala/monocle/optics.scala
```diff
@@ -0,0 +1,197 @@
...
+  def apply[S, T, A, B](modify0: (S, A => B) => T)(set0: (S, B) => T): PSetter[S, T, A, B] =
+    new PSetter[S, T, A, B] {
+      def modify(f: A => B): S => T = modify0(_, f)
+      def set(b: B): S => T = set0(_, b)
+
+    }
+
+}
+
+sealed trait PTraversal[S, T, A, B] extends PSetter[S, T, A, B] { self =>
```
julien-truffaut: PTraversal should extend Fold[S, A] as well

kernel/shared/src/main/scala/monocle/optics.scala
```diff
@@ -0,0 +1,197 @@
...
+                (ez, c) =>
+                  ez.map(f(_, c)) match {
+                    case Left(z) => Left(Left(z))
+                    case Right(a) => a match {
+                      case Left(z) => Left(Left(z))
+                      case Right((z, d)) => Right((Right(z), d))
+                    }
+                  }
+              )(a) match {
+                case (Left(z), b) => ???
```
julien-truffaut: ???

julien-truffaut:
@adamgfraser @jdegoes when you have time, I would love to hear your comments.

kernel/shared/src/main/scala/monocle/optics.scala
```diff
@@ -0,0 +1,197 @@
...
+
+    }
+
+}
+
+sealed trait PTraversal[S, T, A, B] extends PSetter[S, T, A, B] { self =>
+  def traverse[Z](z: Z, f: (Z, A) => Either[Z, (Z, B)]): S => (Z, T)
+
+  final override def fold[Z](zero: Z, combine: (Z, Z) => Z)(f: A => Z)(s: S): Z =
+    traverse[Z](zero, (z, a) => Left(combine(z, f(a))))(s)._1
```
julien-truffaut: I started to play a bit with this encoding. I am not entirely sure what does it mean to return a Left or a Right in traverse. At first, I thought Left was a way to short cut the traversal but it doesn't look to be the case.

@jdegoes @adamgfraser do you have an example of a `Traversal` for all elements in a List.

jdegoes:
@julien-truffaut I've been extraordinarily busy and not had a chance to return to this, but will as soon as I can. Also @adamgfraser and I are working on a radically different encoding that uses one type for everything but we still have work to do.

julien-truffaut:
Closing this as we are making experiment in Monocly reposoitory. But lots of good ideas from @jdegoes and @adamgfraser, thanks.

PR #755, opened 2019-11-26
Fix spelling typos

bwignall:
Should be non-semantic.

Uses https://en.wikipedia.org/wiki/Wikipedia:Lists_of_common_misspellings/For_machines to find likely typos.


julien-truffaut:
Thanks, did you run a program to find them?

bwignall:
I ran the Wikipedia list through a loop of `git grep`s, nothing fully automated.

PR #759, opened 2019-12-12
POC of Fold based on Iterator

julien-truffaut:
It is probably a stupid idea but I wanted to give it a try.

An `Iterator` is mutable and cannot be reused but it is simple and it offers both efficient iteration and laziness.

### DIFF
--- a/kernel/shared/src/main/scala/monocle/Fold.scala
+++ b/kernel/shared/src/main/scala/monocle/Fold.scala
@@ -1,0 +1,64 @@
+package monocle
+
+trait Fold[A, B] { self =>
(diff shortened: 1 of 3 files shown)


cquiroz:
Why do you think it is a stupid idea? Is that due to the use of mutable variables? It seems valid to me to use local mutable state

julien-truffaut:
1. it seems too simple.
1. it returns a mutable object. I guess it is fine, we are used to consuming Iterator only once.

Mainly I am waiting for someone to tell me it is a bad idea but it seems valid to me too.

cquiroz:
Well it seems fine to me 👍

PR #774, opened 2019-12-20
Enforce scalafmt

cquiroz:
This PR enforces checking that scala code is formatted using scalafmt. Scalafmt was also run on the 3.x code


julien-truffaut:
Great work!

PR #778, opened 2019-12-21
Add test for getter syntax

cquiroz:
These maybe unneeded given the Lens test but I though it would be nice to have them


julien-truffaut:
Good idea.

PR #779, opened 2019-12-22
Setter implementation

cquiroz:


core/shared/src/main/scala/monocle/Lens.scala
```diff
@@ -2,7 +2,7 @@ package monocle
 
 import monocle.function.{At, Field1, Field2, Field3, Field4, Field5, Field6}
 
-trait Lens[A, B] extends Optional[A, B] with Getter[A, B] { self =>
+trait Lens[A, B] extends Optional[A, B] with Getter[A, B] with Setter[A, B] { self =>
```
julien-truffaut: I don't think it is necessary here because `Lens` already inherits from `Optional`

core/shared/src/main/scala/monocle/Optional.scala
```diff
@@ -2,7 +2,7 @@ package monocle
 
 import monocle.function.{Cons, Index, Possible}
 
-trait Optional[A, B] extends Fold[A, B] { self =>
+trait Optional[A, B] extends Fold[A, B] with Setter[A, B] { self =>
   def getOption(from: A): Option[B]
   def set(to: B): A => A
```
julien-truffaut: we can probably remove `set` as it inherited from `Setter`

macro/src/main/scala/monocle/macros/syntax/FieldSyntax.scala
```diff
@@ -27,6 +27,10 @@ trait FieldSyntax_Priority2 extends FieldSyntax_Priority3 {
     def field[C](f: B => C): Getter[A, C] = macro FieldSyntaxImpl.field_impl[Getter, A, B, C]
   }
 
+  implicit class GenFieldsSetter[A, B](private val value: Setter[A, B]) {
```
julien-truffaut: should we also add one for `AppliedSetter`?

julien-truffaut:
Awesome, thanks!

PR #780, opened 2019-12-24
Add dotty support

cquiroz:
This is experimental so we may delay merging it if preferred
This PR adds cross compliation to dotty. So far no source code needs to be changed but there are some notes:
* Obviously the macros module doesn't compile
* Tests are not being compiled nor executed as scalatest isn't available. Also I move the `dotSyntax` tests to a `scala-2.13` specific dir
* I removed kind-projector which doesn't seem to be in use
* dotty doesn't work on js


julien-truffaut:
Amazing, thanks a lot @cquiroz 

Just to be sure, is it cross build with dotty in GitHub actions?

cquiroz:
Yes, it behaves like a cross compilation though tests are only compiled and run on 2.13

PR #782, opened 2019-12-28
Check scalafmt directly with sbt

cquiroz:
It is better to use sbt directly to check for format correctness. The action I used earlier is not actively maintained
This version uses sbt and maybe a bit slower but will support upgrades better


julien-truffaut:
LGTM

Issue #771, opened 2019-12-19
Use variance in optic type parameters


julien-truffaut:
A `Getter[A, B]` is equivalent to `A => B`. So the variance of `Getter` should be the same as `Function1`, covariant in A, the input and contravariant in B, the output. If you are like me and have trouble putting your head around variance, you will find a great resource in [Thinking with types](https://leanpub.com/thinking-with-types).

```scala
trait Getter[-A, +B] {
  def get(from: A): B
}
```

On the other hand, a `Lens[A, B]`  is equivalent to a pair of function `(get: A => B, set: (A, B) => A)` . Both A and B appear in covariant and contravariant positions, which means `Lens` must be invariant in A and B. Now, `Lens` inherits `Getter`, so they both need to have the same variance! 

```scala
trait Lens[A, B] extends Getter[A, B] { // variance conflict
  def set(from: A, newValue: B): A
}
```

It looks like we reached a dead end. However, if we look at polymorphic optics, the situation is quite different. A `PolyLens[A1, A2, B1, B2]` is a pair of function `(get: A1 => B1, set: (A1, B2) => A2)`. Now, A1 - B2 are in contravariant position, and A2 - B1 are in covariant position.

```scala
trait PolyLens[-A1, +A2, +B1, -B2] extends Getter[A1, B1] { 
  def set(from: A1, newValue: B2): A2
}
```
 
Thank you, @adamgfraser  and @jdegoes for the idea.

This issue is directly conflicting with https://github.com/julien-truffaut/Monocle/issues/770.

jdegoes:
@julien-truffaut Happy to help!

On miscellaneous other things I got too busy to respond to:

1. **Sealed vs unsealed**. If inheritance is going to be used at all — and I think it _must_ be — then the sealed vs unsealed distinction doesn't really matter that much. After all, if you have a `Fold` you can always type case it to the most powerful variant, e.g. `prism: Prism`. Sealing them is a choice to make that explicit: e.g. saying even if you are passed something weaker, if you match against it, you might find something stronger. The only way to prevent that possibility is to avoid inheritance at all, because only then will you remove the possibility of type casing on the inheritance hierarchy.
2. **Traversal**. The solution that @adamgfraser and I sketched was _not_ as powerful as a full traversal, although it can do many of the same things. If you want the full power, I suggest the simplest possible type class:
    ```scala
    trait Zippable[F[_]] {
      def succeed: F[Unit]
[...]

PR #787, opened 2020-01-02
Read only variance

julien-truffaut:
@joroKr21 pointed out that we can use variance for Getter and Fold even in the monomorphic case.

https://github.com/julien-truffaut/Monocle/issues/771#issuecomment-570255879

### DIFF
--- a/core/shared/src/main/scala/monocle/Fold.scala
+++ b/core/shared/src/main/scala/monocle/Fold.scala
@@ -2,7 +2,7 @@ package monocle
 
 import monocle.function._
 
-trait Fold[A, B] { self =>
+trait Fold[-A, +B] { self =>
   def toIterator(from: A): Iterator[B]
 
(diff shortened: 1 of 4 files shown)


core/shared/src/main/scala/monocle/Fold.scala
```diff
@@ -59,42 +59,42 @@ trait Fold[A, B] { self =>
         self.toIterator(from).flatMap(other.toIterator)
     }
 
-  def asTarget[C](implicit ev: B =:= C): Fold[A, C] =
+  def asTarget[C](implicit ev: B <:< C): Fold[A, C] =
```
joroKr21: Hmm that's interesting, why not `[C >: B]`?

julien-truffaut: yes, it is better!

PR #790, opened 2020-01-07
Friendly type paramaters

julien-truffaut:
This PR renames the type parameters in optics from `A`, `B` to `From` and `To`, e.g.

```scala
trait Lens[From, To] {
  def get(from: From): To
  def set(to: To): From => From
}
```

I believe it should make the code more readable, especially for beginners.

### DIFF
--- a/core/shared/src/main/scala/monocle/Fold.scala
+++ b/core/shared/src/main/scala/monocle/Fold.scala
@@ -1,9 +1,9 @@
 package monocle
 
-trait Fold[A, B] { self =>
-  def toIterator(from: A): Iterator[B]
+trait Fold[From, To] { self =>
+  def toIterator(from: From): Iterator[To]
 
-  def foldLeft[Z](zero: Z)(f: (Z, B) => Z): A => Z =
+  def foldLeft[Z](zero: Z)(f: (Z, To) => Z): From => Z =
     from => {
       var acc = zero
       val it  = toIterator(from)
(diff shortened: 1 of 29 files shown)


diesalbla:
Given that the traditional intuitive approximation to lenses is "A field in a record", would the names `Record` and `Field` suit it?

julien-truffaut:
@diesalbla Yes it will probably help for `Lens` but it is not the case for other optics, e.g. `Prism` or `Iso`. IMO, we should aim for consistent type parameters across all optics. 

However, there is a case for changing the "traditional" names of optics to something like tofu-optics: https://github.com/TinkoffCreditSystems/tofu/blob/master/optics/core/src/main/scala/tofu/optics/classic.scala#L8

cquiroz:
I can see how for a subset of people having the traditional A, B would be easier, however I think it is easier for beginners. I often get confused by the types so for me it is a 👍

cquiroz:
I wonder a better name for X in the `compose` Operations, maybe `Next`, `Last`, `Final` Or `Target`. I admit none of the is very satisfying

diesalbla:
Perhaps `Context` and `Elem` may be a bit more general. You can see a lens as an optic to see the _element_ (the field) inside its context. For a Prism, the _context_ of one element is that it is one alternatives.

julien-truffaut:
@cquiroz I also tried `Next` but I wasn't convinced. I agree `X` is not amazing, what would be your preference?

julien-truffaut:
@diesalbla Thanks for your proposal. To be honest, I don't have a strong opinion either way. Unfortunately, naming is really subjective and asking a poll on twitter will probably not target the beginner audience we are looking for :(

I would suggest to go ahead with the current proposal and re-evaluate when we discuss renaming optics.

PR #793, opened 2020-01-09
stop importing cats.Functor in gen lens macros

### LINKED ISSUE #792: reference to Functor is ambiguous in GenLens macros
`import cats.Functor` can cause name conflict when there's another `Functor` in scope:
https://github.com/julien-truffaut/Monocle/blob/94dfb513d26ed3566ba7374c534a22ed772b83b1/macro/src/main/scala/monocle/macros/internal/Macro.scala#L78

```scala
object X {
  type Functor
 
  @Lenses()
  case class C(i: Int)
}
```

```
reference to Functor is ambiguous;
[error] it is both object X and imported subsequently by 
[error] import cats.Functor
[error] @Lenses
```

enzief:
fix #792

### DIFF
--- a/macro/src/main/scala/monocle/macros/internal/Macro.scala
+++ b/macro/src/main/scala/monocle/macros/internal/Macro.scala
@@ -75,7 +75,6 @@ private[macros] class MacroImpl(val c: blackbox.Context) {
 
     c.Expr[PLens[S, T, A, B]](q"""
       import monocle.PLens
-      import cats.Functor
       import _root_.scala.language.higherKinds // prevent warning at call site
 
       new PLens[$sTpe, $tTpe, $aTpe, $bTpe]{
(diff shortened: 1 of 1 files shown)


julien-truffaut:
Nice, I haven't thought about this issue at all. Do you want us to make a bug release?

enzief:
Yes please, I'm kinda blocked by this. Thank you!

julien-truffaut:
We are having some issues regarding binary compatibility (see https://github.com/julien-truffaut/Monocle/pull/795)

Once we fix it, we will cut 2.0.1. I will let you know when it is ready.

macro/src/main/scala/monocle/macros/internal/Macro.scala
```diff
@@ -85,8 +84,8 @@ private[macros] class MacroImpl(val c: blackbox.Context) {
         override def set(a: $bTpe): $sTpe => $tTpe =
           _.copy($field = a)
 
-        override def modifyF[$F[_]: Functor](f: $aTpe => $F[$bTpe])(s: $sTpe): $F[$tTpe] =
-          Functor[$F].map(f(s.$fieldMethod))(a => s.copy($field = a))
+        override def modifyF[$F[_]: cats.Functor](f: $aTpe => $F[$bTpe])(s: $sTpe): $F[$tTpe] =
+          cats.Functor[$F].map(f(s.$fieldMethod))(a => s.copy($field = a))
```
fthomas: For [macro hygiene](https://docs.scala-lang.org/overviews/quasiquotes/hygiene.html) wouldn't it be better if this were fully qualified, i.e. `_root_.cats.Functor`? I guess the same applies to the import of `monocle.PLens` which should probably be `import _root_.monocle.PLens`.

PR #800, opened 2020-01-14
Rename compose to and then

debian0501:


julien-truffaut:
Thanks a lot @debian0501 

I think those changes were only required for `core`, `dot-syntax` and `macro` modules. The other modules should be deleted in this branch. Anyway, it does hurt.

I noticed a couple of extra white spaces added, you may want to run `sbt scalafmt`

debian0501:
ok, done. Should I still remove the other directories?

julien-truffaut:
Perfect, thanks. No that's ok, we will do in a subsequent PR. It is easier to review the changes.

debian0501:
how are we going to handle this. Should there be also a new issue for change, or should I simply create a new PR?

julien-truffaut:
@debian0501 good point, I created https://github.com/julien-truffaut/Monocle/issues/803

PR #802, opened 2020-01-15
Sbt 1.3.6

pascalh:
Replaces https://github.com/julien-truffaut/Monocle/pull/781

### DIFF
--- a/core/shared/src/main/scala/monocle/Prism.scala
+++ b/core/shared/src/main/scala/monocle/Prism.scala
@@ -318,7 +318,7 @@ sealed abstract class PrismInstances {
   }
 }
 
-final case class PrismSyntax[S, A](private val self: Prism[S, A]) extends AnyVal {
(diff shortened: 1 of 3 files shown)


julien-truffaut:
Thanks, @pascalh.

Why did you transform `case` from syntax classes? Do you know if it is backwards compatible?

pascalh:
The idea was that these classes are only used for internal conversion to enable operator syntax, hence the change should not impact the end user. 
When I raised the PR I was under the assumption that people would not use these classes in their code. I would not merge this in if the assumption turns out to be false though ;)

julien-truffaut:
I also think it doesn't impact end-user but it will break binary compatibility. We can do it but it would be better if we could upgrade sbt without breaking compatibility.

cquiroz:
I don’t get how an sbt upgrade requires a code change

julien-truffaut:
We are on sbt 1.3.10 now.

PR #811, opened 2020-01-27
Add macro hint for Scala 2.13

ybasket:
People who are not familiar with macros might wonder why macro paradise doesn't work on 2.13 and what to do instead, so I think it's worth a little hint.

### DIFF
--- a/docs/src/main/tut/index.md
+++ b/docs/src/main/tut/index.md
@@ -26,10 +26,16 @@ libraryDependencies ++= Seq(
 )
 ```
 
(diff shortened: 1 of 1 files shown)


julien-truffaut:
Great, idea. Thank you

tjheslin1:
@julien-truffaut I had trouble identifying this issue. Wasn't able to find this hint on [the new site](https://www.optics.dev/Monocle/). Would be ideal to make this easy to find there.

julien-truffaut:
@tjheslin1 Good catch. Sorry, I removed it by mistake.

PR #827, opened 2020-03-05
Support scala.js 1.0

cquiroz:
This should publish fine for both 0.6 and 1.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: scala
 
 scala:
 - 2.12.10
-- 2.13.0
+- 2.13.1
 
 jdk:
 - openjdk8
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -2,7 +2,7 @@ language: scala
 
 scala:
 - 2.12.10
-- 2.13.0
```
julien-truffaut: I remember there was an issue with mima and 2.13.1. Maybe it is fixed in new mima version

cquiroz: hmm, that's why it wasn't changed. It seems it worked though

julien-truffaut:
Thanks @cquiroz. Do you want to make the release? If yes, could you run mima beforehand.

PR #830, opened 2020-03-05
Aggregates

cquiroz:
I think the reason publication isn't working properly is because the aggregates are published too


julien-truffaut:
Let's try! Thanks Carlos.

PR #831, opened 2020-03-06
scalafmt

cquiroz:
This PR uses scalafmt to format every source file and adds a check. the configuration is the same as for the 3.x branch


julien-truffaut:
Great, I am sure it will be helpful to all contributors

PR #845, opened 2020-05-06
Use github actions to build

cquiroz:
This replaces travis for github actions. It will build and run tests and handle publication:

Some Notes:
* It runs tests on scala.js 1.0 and 0.6 thus making it a bit slow
* publication should work on merge but remains to be tested, it needs some settings set as secrets
* doc publication has not been tested either
* I added release-drafter, it maybe useful but could be removed if you prefer writing notes manually

As usual with this type of changes it may need several tries to get it right


julien-truffaut:
Thanks @cquiroz , let's merge it and will see :)

I imagine I need to setup `secrets.GITHUB_TOKEN` do you know where is it?

cquiroz:
GITHUB_TOKEN is provided, I'll setup the gpg keys but you need to put credentials for sonatype like
image

julien-truffaut:
done

cquiroz:
Did you do the PGP ones? I don't have access to change secrets

julien-truffaut:
no, I haven't done the PGP ones. I am not sure what they are.

PR #847, opened 2020-05-07
Workaround to build with sbt 1.3.x

cquiroz:
Disables fatal warnings on core to workaround https://github.com/sbt/zinc/issues/688


julien-truffaut:
Thanks @cquiroz !

PR #848, opened 2020-05-08
add publish microsite workflow

julien-truffaut:
I am not sure what should be used in `secrets.REQUIRES_PUBLISH_MICROSITE`


cquiroz:
I guess that variable is not needed

PR #849, opened 2020-05-09
Remove persoanl patreon link, we should look for something for the en…

julien-truffaut:
we should look for something for the entire optics-dev organisation

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,1 +1,0 @@
-patreon: julien_truffaut


cquiroz:
Doesn’t GitHub have a way to support funding?

julien-truffaut:
@cquiroz Yes, Github supports it. There is some paperwork to fill here: https://github.com/sponsors/optics-dev/waitlist

I will try to do apply.

PR #851, opened 2020-05-11
replace tut by mdoc (issue with prism.md)

julien-truffaut:
Locally, I have this weird error:

```
error: Error while emitting prism.md
assertion failed:
  Trying to access the this of another class: tree.symbol = class Session$App16, class symbol = <$anon: monocle.PPrism> compilation unit: prism.md
     while compiling: prism.md
        during phase: jvm
     library version: version 2.13.1
    compiler version: version 2.13.1
...
last tree to typer: TypeTree(class TPrintColors)
       tree position: line 194 of prism.md
            tree tpe: pprint.TPrintColors
              symbol: case class TPrintColors in package pprint
   symbol definition: case class TPrintColors extends Product with Serializable (a ClassSymbol)
      symbol package: pprint
       symbol owners: class TPrintColors
           call site: <$anon: Function1> in package repl in package repl
```

### DIFF
--- a/.github/workflows/scala.yml
+++ b/.github/workflows/scala.yml
@@ -36,7 +36,7 @@ jobs:
           path: ~/.sbt
           key: ${{ runner.os }}-sbt-${{ hashFiles('build.sbt') }}-${{ hashFiles('project/*.scala') }}
           restore-keys: ${{ runner.os }}-sbt-
-      - name: Run tests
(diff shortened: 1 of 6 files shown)


julien-truffaut:
Not sure why `GenIso` caused this weird error. I remove mdoc for that piece of code and it works.

PR #867, opened 2020-06-12
add unsafe Iso[Map[K,V], List([K, V])] and Traversal[Map[K, V], (K, V)]

contrun:
I have a use case in which I need to rename the key of a json object, i.e. rename "_key" to "key".
I am pretty sure that I will not create any key collision. In my use case, creating unsafe instances of `Iso[Map[K,V], List([K, V])]` and `Traversal[Map[K, V], (K, V)]` is useful.

### DIFF
--- a/docs/src/main/mdoc/unsafe_module.md
+++ b/docs/src/main/mdoc/unsafe_module.md
@@ -40,3 +40,43 @@ import monocle.macros.GenLens
 ```
 
 In this example the age is reset to `0` which invalidates the original predicate of `age >= 18`. More formally `UnsafeSelect` can invalidate the `roundTripOtherWayLaw` law.
+
+## MapTraversal
+
+`MapTraversal` provides an `Iso` `allKeyValues` between `Map[K,V]` and `List([K, V])` and Traversal `mapKVTraversal` of `Map[K, V]` to `(K, V)`. They are useful for traversing and modifying the entries of a map.
+
+Both of them are unsafe because of key collision and the unorderness of map entries. As a rule of thumb, laws regarding modifying `(K, V)` and then getting back a `List[(K, V)]` could be broken, while laws regarding modifying `(K, V)` and then getting back a `Map[K, V]` could still hold. For example,
+
+```scala mdoc:silent
+import monocle.unsafe.MapTraversal.allKeyValues
+
+val list = List((1, "foo"), (1, "bar"))
+val list2 = (allKeyValues[Int, String].get _ compose allKeyValues[Int, String].reverseGet)(list)
+```
+
+`list` and `list2` does not equal here due to key collision. Even if there is no key collision, the output list is still not guaranteed to be the original list as the order of map entries may be unspecified.
+
+On the other hand, the composition of `reverseGet` and `get` is identity.
+
+```scala mdoc:silent
+import monocle.unsafe.MapTraversal.allKeyValues
+
+val map = Map(1 -> "foo", 1 -> "bar")
+val map2 = (allKeyValues[Int, String].reverseGet _ compose allKeyValues[Int, String].get)(map)
+```
+
+Here is an example of how to modify the keys of a map with Monocle. Keep in mind creating identical keys can result in surprising behaviour.
+
+```scala mdoc:silent
+import cats.implicits._
+import monocle.Traversal
+import monocle.unsafe.MapTraversal.allKeyValues
+
+val eachL = Traversal.fromTraverse[List, (Int, String)]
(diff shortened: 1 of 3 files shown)


unsafe/src/main/scala/monocle/unsafe/MapTraversal.scala
```diff
@@ -15,6 +15,18 @@ import scala.collection.immutable.Map
 object MapTraversal {
   implicit def mapEach[K, V]: Each[Map[K, V], V] = fromTraverse[Map[K, ?], V]
 
+  implicit def mapTraversal[K, V]: Traversal[Map[K, V], (K, V)] =
```
julien-truffaut, line 18: optics (`Iso`, `Lens`, ...) are not unique per type, so we don't define them with implicit as opposed to typeclass instance.

julien-truffaut: maybe we should use another name like `mapKeyValues`, `allKeyValues`, naming is hard ^^

unsafe/src/main/scala/monocle/unsafe/MapTraversal.scala
```diff
@@ -15,6 +15,18 @@ import scala.collection.immutable.Map
...
+    new Traversal[Map[K, V], (K, V)] {
+      def modifyF[F[_]: Applicative](f: ((K, V)) => F[(K, V)])(s: Map[K, V]): F[Map[K, V]] =
+        s.toList.traverse { case (k, v) => f(k, v) }.map(kvs => Map(kvs: _*))
+    }
+
+  implicit def mapListIso[K, V]: Iso[Map[K, V], List[(K, V)]] =
+    Iso[Map[K, V], List[(K, V)]](_.toList)(Map.from)
+
+  implicit def mapKVTraversal[K, V]: Traversal[Map[K, V], (K, V)] =
```
julien-truffaut, line 27: what's the difference with ` mapTraversal`?

contrun: Well, they are the same thing. Sorry for not looking into the diff carefully before submitting PR.

unsafe/src/main/scala/monocle/unsafe/MapTraversal.scala
contrun: Thanks. I removed the implicit and changed the name to `allKeyValues`.

julien-truffaut:
LGTM, maybe we should add an issue to document why those optics are unsafe and in which case it is safe.

docs/src/main/mdoc/unsafe_module.md
```diff
@@ -40,3 +40,57 @@ import monocle.macros.GenLens
...
+
+`list` and `list2` does not equal here due to key collision. Even if there is no key collision, the output list is still not guaranteed to be the original list as the order of map entries maybe unspecified.
+
+On the other hand, the composition of `reverseGet` and `get` is identity.
+
+```scala mdoc:silent
+import monocle.unsafe.MapTraversal.allKeyValues
+
+val map = Map(1 -> "foo", 1 -> "bar")
+// map: scala.collection.immutable.Map[Int,String] = Map(1 -> foo, 2 -> bar)
```
julien-truffaut, line 66: did you manually add the result? I think this should be produced by mdoc no?

julien-truffaut:
Awesome @contrun, really good documentation! I just made a minor comment regarding the output. Otherwise all good :)

docs/src/main/mdoc/unsafe_module.md
contrun: I am not familiar with mdoc. The current form of the results are obtained from scala repl and some manual work.

julien-truffaut: Yes, I don't think you need any manual work. Locally, you can run `sbt makeMicrosite` to build the site and then use jekyll to see a local version as in: https://47degrees.github.io/sbt-microsites/docs/build-the-microsite/

But I think it is enough to remove the repl comment.

contrun: Thanks for the explanation. I removed the comments.

julien-truffaut:
@contrun unfortunately, we cannot yet see your document on the web site because of https://github.com/optics-dev/Monocle/issues/861 
I need to find some time to look into it. Sorry for the inconvenience and thanks again for your contribution.

cquiroz:
I guess we should do a release with this PR, WDYT?

julien-truffaut:
> I guess we should do a release with this PR, WDYT?

Yes, good idea. Let me push a new tag.

Issue #868, opened 2020-06-19
Add Iso involution


julien-truffaut:
An involution is a method that is its own inverse, e.g. `f(f(x)) == x`

e.g. `List(1,2,3).reverse.reverse == List(1,2,3)`

We should be able to define:

```scala
def involution[A](update: A => A): Iso[A]
```

Bonus point: 
Could we refactor some optics to use `involution`?

PR #869, opened 2020-06-23
Add Iso involuton

mlrv:
I took a quick stab at #868

I wasn't entirely sure whether `involuted` (I decided to pick that as opposed to `involution` to be consistent with [haskell](https://hackage.haskell.org/package/lens-4.19.2/docs/src/Control.Lens.Iso.html#involuted)) should live on the `Iso` object or somewhere else, happy to change it if we think it's better placed elsewhere.

### DIFF
--- a/core/shared/src/main/scala/monocle/Iso.scala
+++ b/core/shared/src/main/scala/monocle/Iso.scala
@@ -336,6 +336,13 @@ object Iso {
   /** alias for [[PIso]] id when S = T and A = B */
   def id[S]: Iso[S, S] =
     PIso.id[S, S]
+
+  /**
+    * create a [[PIso]] where S = T and A = B from
+    * a function that is its own inverse
+    */
+  def involuted[A](update: A => A): Iso[A, A] =
+    Iso(update)(update)
 }
 
 sealed abstract class IsoInstances {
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/monocle/Iso.scala
```diff
@@ -336,6 +336,13 @@ object Iso {
   /** alias for [[PIso]] id when S = T and A = B */
   def id[S]: Iso[S, S] =
     PIso.id[S, S]
+
+  /**
+    * create a [[PIso]] where S = T and A = B from
```
julien-truffaut, line 341: I don't think we need the first line of comment.

julien-truffaut:
Thanks @mlrv LGTM

core/shared/src/main/scala/monocle/Iso.scala
mlrv: As in, is `create a [[PIso]] where S = T and A = B` redundant? I guess we could change the whole comment block to
```
/** create a [[Iso]] from a function that is its own inverse */
```

julien-truffaut: yep perfect. Maybe `an` [[ISO]]

PR #870, opened 2020-06-25
Fix broken class diagram link in docs.

dangerousben:
I'm not sure if this is the right thing to do and it has the obvious disadvantage that it will only work while published, but a PR is quicker to raise than an issue ;)


cquiroz:
@julien-truffaut How can we verify this is right?

julien-truffaut:
@cquiroz we can build and run the microsite locally: https://47degrees.github.io/sbt-microsites/docs/build-the-microsite/

cquiroz:
Tested and it works great, thanks

PR #886, opened 2020-08-26
Add withDefault and atOrElse to replace None by a default value

julien-truffaut:
This is a slightly unlawful `Iso` but it is very convenient when you want to zoom into a Map like structure and `set/modify` the value when the key doesn't exist.


cquiroz:
Should this go into the `unsafe` package?

julien-truffaut:
> Should this go into the `unsafe` package?

Very good question. In theory, yes, but I am tempted to propose to soften our policy with regards to "lawfullness". I mean we should definitely document when an optic or combinator is a bit dodgy and maybe use a prefix/suffix to make it clear, but if it is really useful and the criteria for "lawfullness" are most often satisfied in real-world cases, then I think we should offer it.

That's just my opinion, what do you think? and @yilinwei ?

TimWSpence:
Haskell lens has a similar but slightly different solution: https://hackage.haskell.org/package/lens-4.19.2/docs/Control-Lens-Combinators.html#v:non (if that linking doesn't work properly, the function you are looking for is
`non :: Eq a => a -> Iso' (Maybe a) a `

eg (super useful for nested maps)
```
Map.empty & at "hello" . non Map.empty . at "world" ?~ "!!!"
fromList [("hello",fromList [("world","!!!")])]
```

They similarly note that it's not necessarily lawful (
"Keep in mind this is only a real isomorphism if you treat the domain as being Maybe (a sans v)
This is practically quite useful when you want to have a Map where all the entries should have non-zero values."
)

julien-truffaut:
@TimWSpence correct, I believe I used the name `withDefault` instead of `non`

TimWSpence:
Oh sorry @julien-truffaut, didn't notice that in the implementation! 👍

yilinwei:
Sorry, didn't see this earlier.

I'm personally quite lax about coherence breakages in my own code (I've overridden an `Eq` instance/ `Ord` instances for sorted maps etc...) and I believe the argument being made is that if you consider the equivalence class where `None === default`, then it still follows all the laws.

We should definitely make a note of the problems in the comment - but the any unexpected breakages I suspect is where people will property test the optic rather than anything else. I'm ambivalent on the whether to have a prefix/suffix.

julien-truffaut:
I added tests and documentation for `atOrElse` let me know what you think.

test/shared/src/test/scala/monocle/function/AtSpec.scala
```diff
@@ -1,17 +1,35 @@
...
 
-  implicit def mmapAt[K, V]: At[MMap[K, V], K, Option[V]] = At.fromIso(MMap.toMap)
+  implicit def mmapAt[K, V]: At[MMap[K, V], K, Option[V]] =
+    At.fromIso(MMap.toMap)
 
   checkAll("fromIso", AtTests[MMap[Int, String], Int, Option[String]])
 
   checkAll("ListMap", AtTests[ListMap[Int, String], Int, Option[String]])
+
+  test("atOrElse") {
```
TimWSpence, line 19: Might be nice to have a test for nested maps? Obviously it _should_ just work but given that it's one of the primary motivations for this, it might be nice to test it explicitly

TimWSpence: Actually, this might be better in `AtExample` if that is considered documentation of how to use it. Usage of `at` for nested maps is definitely not completely obvious at first

core/shared/src/main/scala/monocle/function/At.scala
```diff
@@ -21,6 +23,36 @@ abstract class At[S, I, A] extends Serializable {
...
+    * mapDefaultTo0("id1").modify(_ + 1)(counters) == Map("id1" -> 5, "id2" -> 2)
+    * mapDefaultTo0("id3").modify(_ + 1)(counters) == Map("id1" -> 4, "id2" -> 2, "id3" -> 1)
+    * }}}
+    *
+    * `atOrElse`` is a valid Lens only if `defaultValue` is not part of `S`.
+    * For example, `Map("id" -> 0)` breaks the get-set property of Lens:
+    * {{{
+    * val counters = Map("id" -> 0)
+    * val fromGet  = mapDefaultTo0("id").get(counters)   // 0
+    * val afterSet = mapDefaultTo0("id").set(0)(fromGet) // Map()
```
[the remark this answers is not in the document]
TimWSpence: Maybe worth documenting the nested maps usecase here as well?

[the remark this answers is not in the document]
julien-truffaut: No, you're right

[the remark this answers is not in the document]
julien-truffaut: It would be nice to include nested map example but I am afraid it will take lots of space. Maybe somewhere else?

test/shared/src/test/scala/monocle/function/AtSpec.scala
julien-truffaut: Make sense. I am thinking to get rid of `example` module and instead use the website. It is probably easier to access.

core/shared/src/main/scala/monocle/function/At.scala
```diff
@@ -21,6 +23,36 @@ abstract class At[S, I, A] extends Serializable {
 trait AtFunctions {
   def at[S, I, A](i: I)(implicit ev: At[S, I, A]): Lens[S, A] = ev.at(i)
 
+  /**
+    * Creates a Lens that zooms into an index `i` inside `S`.
```
yilinwei, line 27: It really is a pity that there isn't a better way to link documentation in scala. These examples feel as though they belong elsewhere.

yilinwei (approved):
LGTM

PR #893, opened 2020-09-10
Add POptionals into Ior's left and right

chwthewke:
Hello,

This proposal is to add

`theseLeft[A, B, C]: POptional[Ior[A, B], Ior[C, B], A, C]`
and
`theseRight[A, B, C]: POptional[Ior[A, B], Ior[A, C], B, C]`

which can look into the left and right parts of an Ior and modify the value if present.

`TheseSpec` was amended with the appropriate law checks, and I've run the (jvm) tests successfully.

Is this of any interest?

Thanks


core/shared/src/main/scala/monocle/std/These.scala
```diff
@@ -19,6 +20,16 @@ trait TheseOptics {
       case Right(b) => That(b)
     }
 
+  def theseLeft[A, B, C]: POptional[Ior[A, B], Ior[C, B], A, C] =
```
julien-truffaut, line 23: would you mind to rename those `pTheseLeft` and `pTheseRight` and create monophormic aliases e.g.  `theseLeft[A, B] = pTheseLeft[A, B, A]`?
This is because monomoprhic optics have better type inference.

chwthewke: Done!

PR #896, opened 2020-09-11
Lift arbitrary functions to Getters

TimWSpence:
Inspired by Haskell lens' `to :: (Profunctor p, Contravariant f) => (s -> a) -> Optic' p f s a ` which builds a `Getter` from an arbitrary function.


julien-truffaut:
It looks great, thanks Tim.

PR #902, opened 2020-09-23
fix #901 add some to all optics

### LINKED ISSUE #901: Add `some` method to all optics
I think this case is so common that it is worth the cost of a helper method.

Instead of `optic composePrism some`, user could do `optic.some`

What do you think?

julien-truffaut:


cquiroz (approved):
Looks great, we should release this soon

julien-truffaut:
> Looks great, we should release this soon

We also need to fix the release :(

julien-truffaut:
It also doesn't seem to be working for Scala 2.12

julien-truffaut:
`substituteCo` is not available in 2.12, I tried to use `asInstanceOf` but I get the following error for 2.13:

```scala
[error] /Users/julien/project/Monocle/test/shared/src/test/scala/monocle/LensSpec.scala:105:25: Cannot prove that Option[Int] =:= Option[A1].
[error]     obj.applyLens(lens).some.getOption shouldEqual Some(2)
```

So it looks this can only work for Scala 2.13 :(

cquiroz:
We could keep it as a 2.13 only. There are split dirs for 2.12/2.13

julien-truffaut:
> We could keep it as a 2.13 only. There are split dirs for 2.12/2.13

Good idea, but I am not sure how to implement it because it needs to be a method in the optics class.

julien-truffaut:
If there is no easy workaround, I would rather bump the major version and drop Scala 2.12 from the build. WDYT @cquiroz ?
We can always support bug fix the current major version.

cquiroz:
I think it would be fine to drop 2.12, people on 2.12 can still use the older version

PR #908, opened 2020-10-05
Add each shortcut to all optics

julien-truffaut:


cquiroz (approved):
This is looking great. Are you trying to emulate the style Monocle will have in dotty?

julien-truffaut:
> This is looking great. Are you trying to emulate the style Monocle will have in dotty?

Yeah originally I wanted to add those methods to Monocle 3, but I don't know when it will be ready. So I would like to make Monocle 2 more user friendly and simplify the transition to v3 at the same time.

julien-truffaut:
@cquiroz @yilinwei we could use `?` instead of `some` and `*` instead of `each`. I think those symbols are well understood and it will be consistent with the proposed synttax for `Focus` macro. What do you think?

PR #909, opened 2020-10-06
Use ? instead of some and * instead of each

julien-truffaut:
This proposed syntax is similar to XPath and @yilinwei `Focus` macro.

### DIFF
--- a/core/shared/src/main/scala/monocle/Fold.scala
+++ b/core/shared/src/main/scala/monocle/Fold.scala
@@ -86,10 +86,10 @@ abstract class Fold[S, A] extends Serializable { self =>
         s.fold(c => f(Either.left(c)), self.foldMap(a => f(Either.right(a))))
     }
 
-  def each[C](implicit evEach: Each[A, C]): Fold[S, C] =
+  def *[C](implicit evEach: Each[A, C]): Fold[S, C] =
     composeTraversal(evEach.each)
 
-  def some[A1](implicit ev1: A =:= Option[A1]): Fold[S, A1] =
+  def ?[A1](implicit ev1: A =:= Option[A1]): Fold[S, A1] =
     adapt[Option[A1]] composePrism (std.option.pSome)
 
   private def adapt[A1](implicit evA: A =:= A1): Fold[S, A1] =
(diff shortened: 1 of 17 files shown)


cquiroz:
Should we keep both making one the alias of the other? I know some people are very much off put by symbolic operators

julien-truffaut:
> Should we keep both making one the alias of the other? I know some people are very much off put by symbolic operators

I would prefer we offer one or the other, but not both because it is a source of useless conflict in user's code base. Scala 3 is also pushing toward this with the `@alpha` annotation. 

I agree with you, we should aim to limit symbolic operators (I would like to deprecate the optics compose operators for that reason).

`some` and `each` are really common, so it might convenient to have a one letter operator. I also like the analogy to XPath/JsonPath, but this makes those operations more difficult to google. 

I am really on the fence on this one.

yilinwei:
For a new user, the impedance of having symbolic-operators is significant; but once the cost is paid, terseness becomes useful for both writing and readability. That being said, it is easier to remember an word rather than a concatenation of symbols; we should reserve symbolic operators only for the most common of use-cases such as `>>>` and `<<<`. I’ll defer to your judgement if you feel that this use-case is common enough.

In my experience, developers will find reasons to bike-shed regardless of the syntax which is exposed to them — see Wadler’s law. But in general, I don’t  believe having a different syntax for the same task is *necessarily* problematic; I’ve never enforced the rule that you have to use for-comprehension over `flatMap` which would increase uniformity. **Perhaps if we have few symbolic operators but aliases for them, this would be a good compromise?**

In the case of the `Focus` macro, the DSL requires some identifiers which don’t conflict with the field names within an arbitrary case-class. For that reason alone, symbolic operators are preferable in that case — similarly `idx` and `at` are chosen so as not to conflict with existing methods on `List` and `Map`.

julien-truffaut:
> In the case of the Focus macro, the DSL requires some identifiers which don’t conflict with the field names within an arbitrary case-class. For that reason alone, symbolic operators are preferable in that case — similarly idx and at are chosen so as not to conflict with existing methods on List and Map.

Cheers, I didn't think about that.

Let's keep the current syntax, after all `some` and `each` are really short. We can always decide later when we introduce `Focus` if it makes sense to add symbolic operators or not.

PR #912, opened 2020-10-13
Replace atOrElse by withDefault syntax on optics

julien-truffaut:
The two most common operations on optics that point to an `Option` are:
1. focus only on `Some` which we can do with `some`
1. focus on `Some` or replace `None` by some default value

I propose to add `withDefault` method to all optics and apply optics similar to what we've done for `some`.
I removed `atOrElse`, it can easily be achieved by `at(x).withDefault(y)`

Main questions:
1. Is this use case worth the maintenace cost?
1. shall we add syntax for monorphic optics somewhere else? `implicit evTS: T =:= S, evBA: B =:= A, evOpt: A =:= Option[A1]` is not really readable.

### DIFF
--- a/core/shared/src/main/scala/monocle/Fold.scala
+++ b/core/shared/src/main/scala/monocle/Fold.scala
@@ -1,6 +1,6 @@
 package monocle
 
-import cats.{Foldable, Monoid}
+import cats.{Eq, Foldable, Monoid}
 import cats.arrow.Choice
 import cats.instances.int._
 import cats.instances.list._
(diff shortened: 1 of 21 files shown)


yilinwei:
@julien-truffaut Regarding the second point - does IntelliJ filter out such methods?

julien-truffaut:
> @julien-truffaut Regarding the second point - does IntelliJ filter out such methods?

No, IntelliJ offers the method but there is s red squiggly
Screenshot 2020-10-14 at 09 40 25
Screenshot 2020-10-14 at 09 40 44

julien-truffaut:
@yilinwei what do you think of this encoding?

cquiroz:
I like this a lot. It is something I use all the time and would benefit from some syntax improvements.

julien-truffaut:
If you are fine with this, I would prefer to merge this PR first before implementing the same syntax class for `Apply` optics. I prefer to do it in two or more PR as I would like to split each `Apply` optics in a separate class.

yilinwei (approved):

PR #921, opened 2020-10-19
Add scala-steward configuration

julien-truffaut:
Thanks to @fthomas


fthomas (changes requested):
The `.scala-steward.conf` file needs to be in the root directory of the repo.

julien-truffaut:
> The `.scala-steward.conf` file needs to be in the root directory of the repo.

thx

fthomas (approved):

PR #929, opened 2020-10-26
WIP add .lens syntax to Lens and Prism

julien-truffaut:
Add `lens` extension methods to all optics such that we can use `.lens(_.foo)` instead of `composeLens GenLens[Target](_.foo)`


cquiroz (approved):
Looks great!

julien-truffaut:
@yilinwei do you know if it is possible to generalise the implementation so that we don't need to duplicate the macro for each optic type?

yilinwei:
@julien-truffaut 

Yes there should be; did you try returning a `c.Expr[Any]` and invoking the same `impl` with a narrowed type?

If that doesn't work, we can also do an inline method and invoke the macro which should definitely work (you'll need to do a `.cast` with the `Expr`, but that's safe because it still does a typecheck before embedding the tree.

julien-truffaut:
closing, for now, let's review when we have the Focus macro

PR #935, opened 2020-11-03
Creating Iso using the new method "involuted" when it is possible

borisghidaglia:
Completing the "bonus point" of #868 using #869 implementation.

Beware, it's a heavy PR


julien-truffaut:
Thank you @borisghidaglia

PR #937, opened 2020-11-04
Scala 3.0.0

cquiroz:
Updates dotty and tries a fix to the doc issue on release


julien-truffaut:
Thanks let's cross our fingers!

julien-truffaut (approved):

PR #938, opened 2020-11-05
Skip building docs in dotty

cquiroz:
Another attempt, this switches to the latest doc tool for dotty but still fails. Instead we just skip building dotty docs altogether.


julien-truffaut (approved):

julien-truffaut:
Thanks @cquiroz

cquiroz:
It seems this finally worked, it would be nice to be replacing the scala 2 dependencies before making a public release with dotty

PR #942, opened 2020-11-05
Fix links to docs and class-diagram on website

ksonj:
The links on the landing page to the documentation seemed to be broken. I've really no idea what I'm doing here, but locally this appears to fix it. Hope it's okay to send unsolicited PRs. Feel free to close, ignore and reinterpret as a bug report if not :slightly_smiling_face:


julien-truffaut:
Good catch. Thanks, @ksonj.

Let's give it a try, anyway it is already broken.

julien-truffaut:
@ksonj it looks like it works. Thanks a lot.

PR #952, opened 2020-11-16
Fix doc task and skip publishing dotty for now

cquiroz:
This may fix the release failing task.
I'm disabling publishing for dotty, it is recommended not to publish with dependencies from scala 2. Though the code compiles and tests run it may cause dependency issues downstream


julien-truffaut:
Thanks for looking into it, Carlos. It makes sense. The most important point at the moment is to check that it compiles and pass the tests. So that's perfect.

julien-truffaut:
@cquiroz It seems that it worked!

Issue #957, opened 2020-11-16
Can we use singleton type to create At instance for tuples?


julien-truffaut:
For example, could we write:

```scala
val tuple = (1, false, "Hello")

(tuple applyLens at(2)).get == false
(tuple applyLens at(3)).get == "Hello"
```

Here is a library that uses singleton to look for examples https://github.com/travisbrown/sized

nafg:
```scala
import $ivy.`com.github.julien-truffaut::monocle-macro:2.0.3`, monocle.function._, monocle.syntax.all._, monocle.function.At._

class SingletonAtMaterializer[I <: Singleton] { def apply[S, A](implicit ev: At[S, I, A]) = ev}

// I think this could live in AtFunctions as a legal overload
def at[I <: Singleton] = new SingletonAtMaterializer[I]

// implicit conversion
implicit def singletonAtMaterializerToLens[I <: Singleton, S, A](mat: SingletonAtMaterializer[I])(implicit ev: At[S, I, A], I: ValueOf[I]): monocle.Lens[S, A] =
  mat[S, A].at(I.value)

// define lots of these
implicit def at1_tuple2[A, B]: At[(A, B), 1, A] = At[(A, B), 1, A](_ => fields.first[(A, B), A]) 

val tuple = (1, false) 

(tuple applyLens at[1]).get 
```

nafg:
Actually SingletonAtMaterializer#apply wasn't serving any real purpose so...

```scala
import $ivy.`com.github.julien-truffaut::monocle-macro:2.0.3`, monocle.function._, monocle.syntax.all._, monocle.function.At._

class SingletonWrapper[I <: Singleton]

// I think this could live in AtFunctions as a legal overload
def at[I <: Singleton] = new SingletonWrapper[I]

// implicit conversion
implicit def singletonWrapperToLens[I <: Singleton, S, A](wrapper: SingletonWrapper[I])(implicit at: At[S, I, A], I: ValueOf[I]): monocle.Lens[S, A] =
  at.at(I.value)

// define lots of these
implicit def at1_tuple2[A, B]: At[(A, B), 1, A] = At[(A, B), 1, A](_ => fields.first[(A, B), A]) 

val tuple = (1, false) 

(tuple applyLens at[1]).get 
```

nafg:
Alternatively, if a different name is chosen then the implicit conversion can be avoided and it's really simple:

```scala
[...]

PR #959, opened 2020-11-17
Deprecate FieldX for At with a singleton index

julien-truffaut:
Thanks to @nafg for the idea! see #957 

If we like the idea, I would like to:
1. deprecate `First`, `Second`, etc
1. generate At instances for more tuples. Where shall we stop?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -49,7 +49,13 @@ lazy val buildSettings = Seq(
     "-feature",
     "-unchecked",
     "-Xfatal-warnings",
-    "-deprecation"
+    "-deprecation",
+    "-Wconf:msg=class Field1 in package function is deprecated:i",
+    "-Wconf:msg=class Field2 in package function is deprecated:i",
+    "-Wconf:msg=class Field3 in package function is deprecated:i",
+    "-Wconf:msg=class Field4 in package function is deprecated:i",
+    "-Wconf:msg=class Field5 in package function is deprecated:i",
+    "-Wconf:msg=class Field6 in package function is deprecated:i",
   ),
   scalacOptions in (Compile, console) -= "-Ywarn-unused:imports",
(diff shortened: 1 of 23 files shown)


julien-truffaut:
The problem with this approach is that when we use `at("Hello")` the compiler believes that we want to use the singleton type "Hello".

joroKr21:
You could constrain it to `Int with Singleton` but you would still have the same problem ambiguity with `at(2)`

julien-truffaut:
> You could constrain it to `Int with Singleton` but you would still have the same ambiguity problem with `at(2)`

Where would you add this constraint? I guess what we need is a way to say try it with a singleton and if it doesn't work, try with a normal type (not sure about the terminology).

cquiroz (approved):
Looks great 👍

yilinwei:
> Where would you add this constraint? I guess what we need is a way to say try it with a singleton and if it doesn't work, try with a normal type (not sure about the terminology).

@julien-truffaut, did you find a way around this issue?

julien-truffaut:
@yilinwei I didn't but @joroKr21 did! See https://github.com/optics-dev/Monocle/pull/973

yilinwei (approved):

julien-truffaut:
I just thought about something ... What will be the impact for the implicit resolution of `At`? I believe I read the algorithm that find implicit values is `O(n!)` where `n` is the number of instances. If that's the case, is it a bad idea to define all these instances?

joroKr21:
I don't think it's that bad. Most instances will be filtered out by the `S` type parameter. And implicit search is not so slow in newer Scala versions. It's a funny story, the biggest slowdown was coming from creating error messages that were never shown 🙈 but I don't want to dig now for the `scalac` PR that fixed it.

Issue #964, opened 2020-11-19
Add andThen on all optics


julien-truffaut:
Optics and functions compose in the same way

```scala
def compose[A, B, C](f: B => C, g: A => B): A => C
def compose[A, B, C](f: Optic[A, B], g: Optic[B, C]): Optic[A, C]
```

You may have noticed that the parameters of optics composition are the inverse of function composition. Optics composition looks more like `andThen` than `compose`.

```scala
def andThen[A, B, C](f: A => B, g: B => C): A => C
```

Therefore, I propose to create a new overloaded method `andThen` and deprecate all the`composeX` methods later.

For example for lenses will have something like:

```scala
trait PLens[S, T, A, B] {
  def andThen[C, D](other: PLens[A, B, C, D]): PLens[S, T, C, D] = ???
  def andThen[C, D](other: PPrism[A, B, C, D]): POptional[S, T, C, D] = ???
  def andThen[C, D](other: PIso[A, B, C, D]): PLens[S, T, C, D] = ???
  // and so on
}
```

I propose to overload `andThen` instead of using `andThenLens`, `andThenPrism` for a few reasons:

1. it is more concise
1. users don't need to think about what type of optics they are composing
1. the main reason to have non-overloaded compose methods is to help type inference for optics like `at`, `index`, `each` which I think we can get away by adding shortcuts on each optic https://github.com/optics-dev/Monocle/issues/956

If this works well, we can deprecate all `composeLens`, `composePrism`, ... methods.

julien-truffaut:
I have started to work on this

PR #967, opened 2020-11-20
add andThen on all optics

julien-truffaut:
see #964

I believe we can deprecate all `composeX` methods once we have a short cut for `at`, `index` and maybe a few other common optics.


julien-truffaut:
@cquiroz @yilinwei are you ok with this change?

cquiroz (approved):
it seems fine to me, we still have the old names there so it won't break existing code I think

yilinwei:
:+1:

julien-truffaut:
> it seems fine to me, we still have the old names there so it won't break existing code I think

I agree. I would like to eventually deprecate the old names.

PR #971, opened 2020-11-25
cross build two dotty versions

cquiroz:
It has been kind of recommended to build for the last two versions of dotty. This PR adds M2 though we still need to solve some issues to get to publish for dotty


julien-truffaut:
Thank you @cquiroz

julien-truffaut (approved):

PR #973, opened 2020-11-27
Make At contravariant in I so that it works with singletons

joroKr21:
@julien-truffaut that seems to work

The only question - are the other two type parameters of `At` sufficient to avoid problems with implicits and contravariance.

### DIFF
--- a/core/shared/src/main/scala/monocle/function/At.scala
+++ b/core/shared/src/main/scala/monocle/function/At.scala
@@ -13,25 +13,16 @@ import scala.collection.immutable.{ListMap, SortedMap}
 @implicitNotFound(
   "Could not find an instance of At[${S},${I},${A}], please check Monocle instance location policy to " + "find out which import is necessary"
 )
-abstract class At[S, I, A] extends Serializable {
-  def at(i: I): Lens[S, A]
-}
-
-abstract class AtSingleton[S, I <: Singleton, A] extends At[S, I, A] {
+abstract class At[S, -I, A] extends Serializable {
   def at(i: I): Lens[S, A]
 }
 
 trait AtFunctions {
-  def at[S, I, A](i: I)(implicit ev: At[S, I, A]): Lens[S, A] = ev.at(i)
+  def at[S, I, A](i: I)(implicit ev: At[S, i.type, A]): Lens[S, A] = ev.at(i)
 
   /** delete a value associated with a key in a Map-like container */
   def remove[S, I, A](i: I)(s: S)(implicit ev: At[S, I, Option[A]]): S =
     ev.at(i).set(None)(s)
-}
-
-trait AtSingletonFunctions {
-  def at[S, I <: Singleton, A](i: I)(implicit ev: AtSingleton[S, I, A]): Lens[S, A] =
-    ev.at(i)
 }
 
 object At extends AtFunctions {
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/monocle/function/At.scala
```diff
@@ -13,27 +13,18 @@ import scala.collection.immutable.{ListMap, SortedMap}
...
-}
-
-abstract class AtSingleton[S, I <: Singleton, A] extends At[S, I, A] {
+abstract class At[S, -I, A] extends Serializable {
   def at(i: I): Lens[S, A]
 }
 
 trait AtFunctions {
-  def at[S, I, A](i: I)(implicit ev: At[S, I, A]): Lens[S, A] = ev.at(i)
+  def at[S, I, A](i: I)(implicit ev: At[S, i.type, A]): Lens[S, A] = ev.at(i)
```
julien-truffaut, line 21: oh, that's interesting. If you have time, would you mind to explain me why you use `i.type` here?

julien-truffaut:
@joroKr21 That's a fantastic idea! 

What are the "problems with implicits and contravariance" that you mentioned? We have quite a few examples of `at` usage in the codebase, it is good sign the test pass.

core/shared/src/main/scala/monocle/function/At.scala
joroKr21: For primitive types it has the same effect as `<: Singleton` (prevents widening) and for reference types it's exactly what it looks like - the singleton type of the parameter. But the cool thing is that it works in both cases with the same method. Only we need `I` to be contravariant so that `At[S, i.type, A]` is also eligible for `At[S, I, A]`.

joroKr21:
The problem is that in Scala 2 the more specific implicit is determined in part by subtyping so if we have both `At[Map[String, Int], String, Option[Int]]` and `At[Map[String, Int], CharSequence, Option[Int]]` the latter will be more specific because `String <: CharSequence` and due to contravariance `At[Map[String, In], CharSequence, Option[Int]] <: At[Map[String, Int], String, Option[Int]]`. I don't know how common is for users to define their own `At` instances. But I think the other two (invariant) type parameters should be enough to mitigate the problem. But for a type class with only one contravariant type parameter (e.g. `Eq[-A]`) that is catastrophic because `Eq[Any]` becomes the most specific instance.

core/shared/src/main/scala/monocle/function/At.scala
julien-truffaut: ah I see, thank you!

julien-truffaut (approved):

julien-truffaut:
> The problem is that in Scala 2 the more specific implicit is determined in part by subtyping so if we have both `At[Map[String, Int], String, Option[Int]]` and `At[Map[String, Int], CharSequence, Option[Int]]` the latter will be more specific because `String <: CharSequence` and due to contravariance `At[Map[String, In], CharSequence, Option[Int]] <: At[Map[String, Int], String, Option[Int]]`. I don't know how common is for users to define their own `At` instances. But I think the other two (invariant) type parameters should be enough to mitigate the problem. But for a type class with only one contravariant type parameter (e.g. `Eq[-A]`) that is catastrophic because `Eq[Any]` becomes the most specific instance.

I understand now. I think it should be fine. We'll publish a couple of milestone releases for people to try. Thanks for your help!

PR #974, opened 2020-11-28
Replace deprecated set with replace

### LINKED ISSUE #961: Deprecate set method and add replace
`set` is a source of confusion because there is an implicit expectation it will insert data if it is missing, for example:

```scala
case class User(name: String, email: Option[String])

GenLens[User](_.email).some.set("j@foo.com")(User("John", None))
// res: User = User("John", None)
```

While most people expect `set` to return `User("John", Some("j@foo.com"))`.

I believe `replace` doesn't have the drawback.

1. deprecate `set` on all optics and apply optics that have it
1. add `replace` as an alias to set or the other way around (set as an alias to replace)
1. use `replace` on all tests and documentation
1. bonus: define scalafix rule

sapizhak:
Closes #961
@julien-truffaut


julien-truffaut:
Thank you very much @sapizhak 
I just updated the `since` tag for the most likely nest release.

sapizhak:
@julien-truffaut great, thanks for changing the since block for the right one - that was initially the thing I wanted to ask you about, i mean the next version :))

julien-truffaut:
I am not sure to understand why the build failed. It doesn't seem related to this change.

julien-truffaut (approved):

julien-truffaut:
Thanks Bogdan

Issue #956, opened 2020-11-16
Add shortcut for at


julien-truffaut:
Add an extension method for all optics (`Lens`, `Iso`, ...) and apply optics (`ApplyLens`, `ApplyIso`) for `at` so that we can do:

```scala
optic.at(10)
```

instead of

```scala
optic composeLens at(10)
```

This should be very similar to what we have done for `each`:
https://github.com/optics-dev/Monocle/blob/master/core/shared/src/main/scala/monocle/Lens.scala#L284
https://github.com/optics-dev/Monocle/pull/908/files (but we moved it to implicit class)

Please write tests to check type inference is working fine (see https://github.com/optics-dev/Monocle/blob/master/test/shared/src/test/scala/monocle/LensSpec.scala#L121-L129)

sapizhak:
i have started working on this one

PR #978, opened 2020-12-05
Add shortcuts for at, index

### LINKED ISSUE #955: Add shortcut for index
Add an extension method for all optics (`Lens`, `Iso`, ...) and apply optics (`ApplyLens`, `ApplyIso`) for `index` so that we can do:

```scala
optic.index(10)
```

instead of

```scala
optic composeOptional index(10)
```

This should be very similar to what we have done for `each`:
https://github.com/optics-dev/Monocle/blob/master/core/shared/src/main/scala/monocle/Lens.scala#L284
https://github.com/optics-dev/Monocle/pull/908/files (but we moved it to implicit class)

Please write tests to check type inference is working fine (see https://github.com/optics-dev/Monocle/blob/master/test/shared/src/test/scala/monocle/LensSpec.scala#L121-L129)

sapizhak:
@julien-truffaut 
Closes #955, #956

### DIFF
--- a/core/shared/src/main/scala/monocle/syntax/ApplyFold.scala
+++ b/core/shared/src/main/scala/monocle/syntax/ApplyFold.scala
@@ -17,22 +17,10 @@ case class ApplyFold[S, A](s: S, _fold: Fold[S, A]) {
   @inline def isEmpty: Boolean                      = _fold.isEmpty(s)
   @inline def nonEmpty: Boolean                     = _fold.nonEmpty(s)
 
-  def each[C](implicit evEach: Each[A, C]): ApplyFold[S, C] =
-    composeTraversal(evEach.each)
-
   def some[A1](implicit ev1: A =:= Option[A1]): ApplyFold[S, A1] =
     adapt[Option[A1]] composePrism (std.option.pSome)
 
-  def withDefault[A1: Eq](defaultValue: A1)(implicit ev1: A =:= Option[A1]): ApplyFold[S, A1] =
-    adapt[Option[A1]] composeIso (std.option.withDefault(defaultValue))
-
-  def at[I, A1](i: I)(implicit evAt: At[A, i.type, A1]): ApplyFold[S, A1] =
-    composeLens(evAt.at(i))
-
-  def index[I, A1](i: I)(implicit evIndex: Index[A, I, A1]): ApplyFold[S, A1] =
-    composeOptional(evIndex.index(i))
-
-  private def adapt[A1](implicit evA: A =:= A1): ApplyFold[S, A1] =
+  private[monocle] def adapt[A1](implicit evA: A =:= A1): ApplyFold[S, A1] =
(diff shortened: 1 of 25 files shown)


core/shared/src/main/scala/monocle/syntax/ApplyFold.scala
```diff
@@ -67,3 +61,22 @@ case class ApplyFold[S, A](s: S, _fold: Fold[S, A]) {
   /** alias to composeIso */
   @inline def ^<->[B, C, D](other: PIso[A, B, C, D]): ApplyFold[S, C] = andThen(other)
 }
+
+object ApplyFold {
+  implicit def applyFoldSyntax[S, A](self: ApplyFold[S, A]): ApplyFoldSyntax[S, A] =
+    new ApplyFoldSyntax(self)
+}
+
+final case class ApplyFoldSyntax[S, A](self: ApplyFold[S, A]) extends AnyVal {
```
julien-truffaut, line 70: what are the benefits of defining `ApplyFoldSyntax`? `Fold` and `Getter` are monomorphic (2 type parameters instead of 4), so I am not sure we need these syntax classes.

julien-truffaut:
Thanks @sapizhak Could you also add a test to check `at` works with tuples? My guess it that we need to make `I` contravariant (i.e. `def at[-I, A1]`)

core/shared/src/main/scala/monocle/syntax/ApplyFold.scala
sapizhak: yes, you are right, as they are monomorphic, that does not bring any benefits except for consistency with the neighboring objects. i'll revert the change, it was mostly experimental :))

sapizhak:
Great idea, I have added some tests to cover all the `At` and `Index` summoned by `at` and `index` functions (probably every instance is called implicitly now in tests, so if a new instance is added, it would be fine to be called in tests too). 

As for contravariant position for `I`.. We cannot set variance annotations for functions in their type parameters, because variance is a quality of classes, whereas functions declared as `def` only implement `FunctionN` trait after desugaring. 

Apart from that, `Function1 ... FunctionN` type parameters are already variance annotated: their arguments are _contravariant_ and result type is _invariant_. An implicit argument scenario would extend a trait where arguments (including an implicit one) are contravariant as well:
```
trait ImplicitFunction2[-A, -AA, +B] {
    def apply(a: A)(implicit i: AA): B
  }
```

So an `I` should already be contravariant! :))

julien-truffaut:
> As for contravariant position for `I`.. We cannot set variance annotations for functions in their type parameters, because variance is a quality of classes, whereas functions declared as `def` only implement `FunctionN` trait after desugaring.

You are completely right, good point.

> Great idea, I have added some tests to cover all the At and Index summoned by at and index functions (probably every instance is called implicitly now in tests, so if a new instance is added, it would be fine to be called in tests too).

Cheers, that's a lot of work but at least we are covered :)

PR #982, opened 2020-12-08
refined to compile to dotty and publish dotty packages

cquiroz:
Lets refined to be compiled with dotty and I removed the restriction to publish on dotty as we don't have dependencies requiring `withDottyCompat`


build.sbt
```diff
@@ -397,9 +399,3 @@ lazy val noPublishSettings = Seq(
   publishArtifact := false,
   skip in publish := true
 )
-
```
cquiroz, line 400: I think this is not needed anymore

julien-truffaut: Yes, you are right

julien-truffaut:
Thanks Carlos!

julien-truffaut (approved):

PR #985, opened 2020-12-09
Macro Perf Test using Probably

### LINKED ISSUE #965: Setup a way to measure macro performance
@propensive created a library call (probably)[https://github.com/propensive/probably] that can measure how long it takes to generate macros.

It woould be useful to measure the performance of `GenLens` and compare it to the new `Focus` macro https://github.com/optics-dev/Monocle/issues/954.

See some discussion on Twitter: https://twitter.com/propensive/status/1329745251679035394

asjad02:
fixes #965 

image

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -108,6 +108,7 @@ lazy val refinedScalacheck = Def.setting("eu.timepit" %%% "refined-scalacheck" %
 lazy val discipline      = Def.setting("org.typelevel" %%% "discipline-core" % "1.1.2")
 lazy val munit           = Def.setting("org.scalameta" %% "munit" % "0.7.16" % Test)
 lazy val munitDiscipline = Def.setting("org.typelevel" %% "discipline-munit" % "1.0.3" % Test)
+lazy val probably        = Def.setting("com.propensive" %% "probably-cli" % "0.8.0")
 
 lazy val macroVersion = "2.1.1"
 
@@ -226,7 +227,8 @@ lazy val macros = crossProject(JVMPlatform, JSPlatform)
     scalacOptions += "-language:experimental.macros",
     libraryDependencies ++= Seq(
       scalaOrganization.value % "scala-reflect"  % scalaVersion.value,
-      scalaOrganization.value % "scala-compiler" % scalaVersion.value % "provided"
+      scalaOrganization.value % "scala-compiler" % scalaVersion.value % "provided",
+      probably.value
     ),
     unmanagedSourceDirectories in Compile += (sourceDirectory in Compile).value / s"scala-${scalaBinaryVersion.value}"
   )
--- a/macro/src/main/scala/monocle/macros/GenPrism.scala
+++ b/macro/src/main/scala/monocle/macros/GenPrism.scala
@@ -12,16 +12,14 @@ object GenPrism {
 }
 
 private class GenPrismImpl(val c: blackbox.Context) {
-  def genPrism_impl[S: c.WeakTypeTag, A: c.WeakTypeTag]: c.Expr[Prism[S, A]] = {
-
+  def genPrism_impl[S: c.WeakTypeTag, A: c.WeakTypeTag]: c.Expr[Prism[S, A]] =
     c.test(s"constructing Prism with S=${c.weakTypeOf[S]}, A=${c.weakTypeOf[A]}") {
       import c.universe._
 
       val (sTpe, aTpe) = (weakTypeOf[S], weakTypeOf[A])
 
       val sTpeSym = sTpe.typeSymbol.companion
-      c.Expr[Prism[S, A]](
-        q"""
+      c.Expr[Prism[S, A]](q"""
       import monocle.Prism
       import scala.{Either => \/, Right => \/-, Left => -\/}
 
@@ -39,5 +37,4 @@ private class GenPrismImpl(val c: blackbox.Context) {
       }
     """)
     }.check(_ => true)
-  }
 }
--- a/macro/src/main/scala/monocle/macros/GenPrism.scala
+++ b/macro/src/main/scala/monocle/macros/GenPrism.scala
@@ -1,6 +1,7 @@
 package monocle.macros
 
 import monocle.Prism
+import probably._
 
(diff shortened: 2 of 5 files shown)


build.sbt
```diff
@@ -108,6 +108,7 @@ lazy val refinedScalacheck = Def.setting("eu.timepit" %%% "refined-scalacheck" %
 lazy val discipline      = Def.setting("org.typelevel" %%% "discipline-core" % "1.1.2")
 lazy val munit           = Def.setting("org.scalameta" %% "munit" % "0.7.16" % Test)
 lazy val munitDiscipline = Def.setting("org.typelevel" %% "discipline-munit" % "1.0.3" % Test)
+lazy val probably        = Def.setting("com.propensive" %% "probably-cli" % "0.8.0")
```
julien-truffaut, line 111: should we enable this dependency only for tests? e.g. `% Test`

julien-truffaut:
Thanks a lot @asjad02
I am a bit surprised by the second result , it says it tooks 0 sec to generate `GenLens[Employee](_.company.address.street.number)`

\cc @propensive

julien-truffaut:
Btw the build fail for the following reason:
```
[error] /home/runner/work/Monocle/Monocle/macro/src/test/scala/monocle/macros/MacroTests.scala:30:12: discarded non-Unit value
[error]     }.check(_ => true)
[error]            ^
[error] one error found
```

I am not sure what `check` returns.

propensive:
`check` returns the value of body of the test (in theory making a test a referentially-transparent wrapper around an expression). You will want to change `check` to `assert` and it will return a `Unit` instead.

propensive:
I'm not sure how much logic is being done in the `0.000`-timed result. But it may be correct if most of the work in the compiler is done in typechecking the result of the macro. Unfortunately that can't be timed in the macro itself. If you can have the same macro invoked multiple times with the same name, then the results will be aggregated and the min/avg/max times shown, which might give a bit more information.

We're also going to move to using `nanoTime` soon to get a bit more precision.

build.sbt
propensive: Yes, I think so. (But try it to make sure...)

macro/src/test/scala/monocle/macros/MacroTests.scala
```diff
@@ -0,0 +1,32 @@
...
+object MacroTests extends Suite("Monocle Macro Tests") {
+  def run(test: Runner): Unit = {
+
+    test("construct Lens[Employee, Address] using GenLens") {
+      GenLens[Employee](_.company.address)
+    }.check(_ => true)
+
+    test("construct Lens[Employee, Int] using GenLens") {
+      GenLens[Employee](_.company.address.street.number)
+    }.check(_ => true)
```
propensive, line 22: You could prefix this with `for(i <- 1 to 100)` to run the test a hundred times, if you wanted to see how consistent it is.

propensive:
Actually, I've just realised that you're not actually testing the *compile-time* performance of Monocle! This is testing the runtime performance of constructing the trees generated by the macro. For macro performance, you'll need to edit the macro code, wrapping the macro implementation in a compile-time test.

To do that, just import `probably._` and use `ctx.test(<some test name>) { /* the macro */ }.check(_ => true)` where `ctx` is the black/whitebox context. You can name the test whatever you like, but it makes sense to generate a name in the macro based on what it's generating. If a test is run more than once with the same name, the results just get aggregated.

The results will be printed to `stdout` automatically after compilation.

julien-truffaut:
> To do that, just import `probably._` and use `ctx.test(<some test name>) { /* the macro */ }.check(_ => true)` where `ctx` is the black/whitebox context. You can name the test whatever you like, but it makes sense to generate a name in the macro based on what it's generating. If a test is run more than once with the same name, the results just get aggregated.

Hi Jon, thanks for the comments! However, I am not sure what you mean by `ctx` is the blackbox context. I only know how to get it when I define a macro e.g. https://github.com/optics-dev/Monocle/blob/master/macro/src/main/scala/monocle/macros/internal/Macro.scala#L11

but I am not sure how to get hold of this  context within a test. By any chance do you have an example? I tried to find one in `probably` without luck.

propensive:
Hi Julien, I probably wasn't clear enough about how this works! In the macro where you have `ctx` available is _exactly_ where you need to define the Probably test: it's the macro definition code that needs to have the Probably test added, because we're testing the performance of the evaluation of that macro code during the runtime _of the compiler_ (so, at "compiletime" by the normal definition).

The things we're testing the performance of would be written exactly the same as they usually are, without reference to Probably, and the test would run every time the macro is expanded. However, the test names are generated dynamically, so you can use the `ctx` value to choose meaningful names for each test. That could be the compilation unit or some other string representation derived from the context.

It's worth saying that you would likely want some way to turn Probably's testing on or off. In Magnolia, I do it with an annotation (`@performance`) which, if Magnolia sees it anywhere, turns on performance testing for the entire compilation.

macro/src/main/scala/monocle/macros/GenPrism.scala
```diff
@@ -12,12 +13,15 @@ object GenPrism {
 
 private class GenPrismImpl(val c: blackbox.Context) {
   def genPrism_impl[S: c.WeakTypeTag, A: c.WeakTypeTag]: c.Expr[Prism[S, A]] = {
-    import c.universe._
 
-    val (sTpe, aTpe) = (weakTypeOf[S], weakTypeOf[A])
+    c.test(s"constructing Prism with S=${c.weakTypeOf[S]}, A=${c.weakTypeOf[A]}") {
```
julien-truffaut, line 17: Oh I see what you mean, so the probably test need to be written in the code of the macro. That's a bit annoying but I understand.

If that's the case, I guess the simplest option would be to keep this code in a separate branch soo that we can run the benchmarks but it will not affect the users. What do you think?

asjad02:
Thanks a lot, @propensive  for #991 

Results::
image

julien-truffaut:
Awesome, great work @asjad02 and @propensive 

I suggested we keep this in a separate branch so that we don't add the probably dependency to macro. What do you think?

PR #991, opened 2020-12-16
Proof of concept using Probably to test macro performance

propensive:
@julien-truffaut Here's a proof of concept of the changes that are needed to run Probably at compiletime to check the macro performance. I didn't pay any attention to _what_ I was testing, but if you run a compilation which expands a macro, then Probably's results will be printed at the end of compilation.


julien-truffaut:
Thank you @propensive

julien-truffaut:
Thanks Jon, I merged https://github.com/optics-dev/Monocle/pull/985

PR #998, opened 2020-12-22
move Map instance for Each and FilterIndex from unsafe module to core

julien-truffaut:


cquiroz (approved):
I use these quite a bit and welcome not needing to include an `unsafe` module to use them

PR #1000, opened 2020-12-23
deprecate all composeX methods in favor of andThen

julien-truffaut:


cquiroz (approved):
LGTM
I should prepare my sed and vim skills to regex replace my uses of compose 😀

julien-truffaut:
> LGTM
> I should prepare my sed and vim skills to regex replace my uses of compose 😀

Indeed, but I think we need to start looking at scalafix to ease up the migration

cquiroz:
Yeah a scalafix would be great for 3.0

PR #1003, opened 2020-12-24
deprecate all applyX methods and operators in favor of optics method

julien-truffaut:
1. deprecate `applyX`, e.g. `applyLens`, `applyGetter`
2. deprecate symbolic aliases to `applyX` e.g. `&<->`
3. add `optics` method which transaform any value `a: A` into a `ApplyIso[A, A, A , A]` i.e. a partlially applied optics that points to itself

What do you think of `optics` name? it needs to be short.

### DIFF
--- a/core/shared/src/main/scala/monocle/syntax/Apply.scala
+++ b/core/shared/src/main/scala/monocle/syntax/Apply.scala
@@ -5,6 +5,8 @@ import monocle._
 object apply extends ApplySyntax
 
 trait ApplySyntax {
+  implicit def toApplyOpticsOps[S](value: S): ApplyOpticsOps[S] = ApplyOpticsOps(value)
+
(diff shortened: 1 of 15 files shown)


core/shared/src/main/scala/monocle/syntax/Apply.scala
```diff
@@ -23,62 +25,79 @@ trait ApplySyntax {
     new ApplyTraversalOps(value)
 }
 
+final case class ApplyOpticsOps[S](private val s: S) extends AnyVal {
+  def optics[T, A, B]: ApplyIso[S, S, S, S] = ApplyIso(s, Iso.id)
```
cquiroz, line 29: I think it is fine

cquiroz (approved):
LGTM

PR #1012, opened 2021-01-20
remove benchmarks against Scalaz and Shapeless optics

julien-truffaut:
These are not popular optics implementation. I guess we should more compare to tofu and quick-lens.


cquiroz (approved):
LGTM

PR #1018, opened 2021-02-03
Update build

julien-truffaut:
* bumped to latest scala 3.0.0 version (M3), I removed cross-compile to M2, refined wasn't available
* removed `-source:3.0-migration` scalac option from `code` because the `Focus` macro used new syntax. Can we make it specific to `Focus` macro?
* I tried to cross-compile test modules but I got some errors regarding mixed dependency (2.13/3.0.0)
* ported `Focus`  macro developed by @kenbot and @yilinwei in Monocly

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -42,7 +42,7 @@ lazy val buildSettings = Seq(
   scalacOptions in (Compile, console) -= "-Ywarn-unused:imports",
   scalacOptions ++= {
     if (isDotty.value)
-      Seq("-Ykind-projector", "-language:implicitConversions,higherKinds,postfixOps")
+      Seq("-source:3.0-migration", "-Ykind-projector", "-language:implicitConversions,higherKinds,postfixOps")
     else Seq(
       "-Ymacro-annotations",
       "-Ywarn-dead-code",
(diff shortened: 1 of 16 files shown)


build.sbt
```diff
@@ -94,24 +82,24 @@ lazy val buildSettings = Seq(
   Compile / doc / sources := { if (isDotty.value) Seq() else (Compile / doc / sources).value }
 )
 
-lazy val catsVersion  = "2.3.0"
-lazy val dottyVersions = Seq("3.0.0-M1", "3.0.0-M2")
+lazy val catsVersion  = "2.3.1"
+lazy val dottyVersions = Seq("3.0.0-M3")
```
cquiroz, line 86: Any reason not to keep both `M2` and `M3` it has been suggested to keep the latest two versions to simplify releasing.

core/shared/src/main/scala-3.x/monocle/Focus.scala
```diff
@@ -0,0 +1,17 @@
+package monocle
+
+import monocle.internal.focus.FocusImpl
+
+object Focus {
+
+  extension [A] (opt: Option[A])
+    def some: A = scala.sys.error("Extension method 'some' should only be used within the moocle.Focus macro.")
+
```
cquiroz, line 9: extra line

build.sbt
julien-truffaut: I tried but it seems that refined doesn't publish M2 for the latest version.

cquiroz: Ok, the two versions idea should be used only when it doesn't create difficulties

cquiroz (approved):
Looks good though I can't really comment on the macro

julien-truffaut:
@kenbot good to know, that settles it :)

Issue #1020, opened 2021-02-03
OrElse on Optional


julien-truffaut:
Try one Optional, if it doesn’t work, try the next.

```scala
sealed trait User

case class Editor(. id: Int, favoriteFont: String)  extends User
case class Reader(id: Int, isPremium: Boolean) extends User

val idEditor: Optional[User, Int] = GenPrism[User, Editor] andThen GenLens[Editor](_.id)
val idReader: Optional[User, Int] = GenPrism[User, Reader] andThen GenLens[Reader](_.id)

val id: Optional[User, Int] = idEditor.orElse(idReader)

id.replace(5)(Editor(1, “Comic Sans”))
// res: User = Editor(5, “Comic Sans”)

id.replace(5)(Reader(1, false))
// res: User = Reader(5, false)
```

Is it a valid Optional? If not, we can add it but we need a document when it deviates from the standard behaviour.

kenbot:
I think this is a valid analogy with `scala.Option.orElse`, and should strengthen users' intuition about the library.

This also makes a decisive case for `getOrElse` instead of `withDefault`

julien-truffaut:
Implemented in #1021

PR #1021, opened 2021-02-03
add orElse on Optional

julien-truffaut:


julien-truffaut:
fix #1020

julien-truffaut:
I think I found an example where this would break the law. I will add some doc next week.

julien-truffaut:
Actually no, the example I had in mind is lawful. 

@kenbot @cquiroz would you mind to review this

cquiroz (approved):
Looks great

PR #1037, opened 2021-02-08
Add GenLens for Scala 3, redirect to Focus

julien-truffaut:
/cc @asjad02

## Macros
* split macros module into Scala 2 et Scala 3 files
* Implement `GenLens` for Scala 3 using `Focus` https://github.com/optics-dev/Monocle/blob/61ceedb076ec5d5e163ee1165ad49b7fe108d94a/macro/src/main/scala-3.x/monocle.macros/GenLens.scala#L6-L8
* Move all macro tests to macros modules

## Generic
* move generic tests to generic module

## Tests
* enable Scala 3
* remove all usage of `GenX` in test module
* remove shapeless
* disable `StringsSpec`, there was an issue with ScalaCheck and refined (I will investigate in a separate ticket)

## Todo later
* re-enable `StringsSpec`
* Implement `GenPrism` using `as` from `Focus` https://github.com/optics-dev/Monocle/issues/1028
* Implement `GenIso` problably tricky

The alternative for macros is to wait for Scala 2.13.5 which should allow to cross compile macro to Dotty. However, I am bit worried how our build will handle it (2.13, 3.0 and 2.13 compiling to 3.0).


kenbot (approved):
Lgtm. 

I think at some point we will have to revisit the `GenXXX` and `Focus` tests and unify them into something that makes sense. But it doesn't have to happen in this PR, I think, it can come later.

Issue #1027, opened 2021-02-08
[Focus] Single-use with pre-applied target


julien-truffaut:
Julien had an idea:

> 
> It is probably more common to create an optic and use it only once, especially for a new user of the library. It would be great if we could make the Monocle API so attractive that users would reach out for it if they have 2 nested copies or even the one.
> 
> The following syntax could tick all the boxes: `foo.focus(_.field1.?.field2.at("x")).get/replace/modify`. It would be equivalent to do:
> ```
> val optic = Focus[Foo](_.field1.?.field2.at("x"))
> optic.get(foo)
> ```
> Another benefit is that you don't have to name the optic. Naming is hard and there is no standard naming convention for optics.
> 

It should be easy enough to spike up. 

Original ticket: https://github.com/optics-dev/Monocly/issues/24

PR #1039, opened 2021-02-09
Focus macro pre-applied to object

kenbot:
Adds the fluent pre-applied syntax to the `Focus` macro, ie `address.focus(_.streetName)`, resolving #1027.

The solution uses the existing `monocle.syntax.ApplyXXX` classes.


julien-truffaut (approved):
Awesome, great job!

Issue #1028, opened 2021-02-08
[Focus] add `as` for enumerations


julien-truffaut:
```scala
sealed trait User {
  def id: Long
}

case class Reader(id: Long, isPremium: Boolean) extends User
case class Editor(id: Long, favoriteFont: String) extends User

Focus[User](_.as[Reader])
// res: Prism[User, Reader]

Focus[User](_.as[Reader].isPremium)
// res: Optional[User, Boolean]
```

Operator to generate a Prism based on a type-cast.

I'm pretty sure that any type-cast gives us a Prism, so we shouldn't need to limit it to sealed hierarchies or enums.

kenbot:
Given that this is a useful enough feature to have a privileged syntax in Focus, it's worth considering whether it should also be a Monocle operator satisfying `Focus[Foo](_.bar.as[Baz]) == Focus[Foo](_.bar).as[Baz]`

PR #1050, opened 2021-02-11
[Focus] `as` feature

kenbot:
"As" feature in Focus, resolving #1028 .

Also some minor cleanup and error message improvements thrown in.  I tried to trip it up with an existential type (ie `Box[_]`), but it turned out that it was the FieldSelect feature that was really choking on this, so I've added a commented-out failing test for the FieldSelectTest.

I had to implement the `as` Prism by hand in the `CastAsGenerator` file; since it's useful enough to have privileged syntax inside Focus, it's worth thinking about whether to also add this as a standalone operator satisfying `Focus[Foo](_.bar.as[Baz]) == Focus[Foo](_.bar).as[Baz]`, then I can just reuse that one inside Focus.

If we can think of more error conditions and ways to trip it up, we can add these later.

### DIFF
--- a/core/shared/src/main/scala-3.x/monocle/internal/focus/features/castas/CastAsParser.scala
+++ b/core/shared/src/main/scala-3.x/monocle/internal/focus/features/castas/CastAsParser.scala
@@ -11,10 +11,14 @@ private[focus] trait CastAsParser {
 
     def unapply(term: Term): Option[FocusResult[(Term, FocusAction)]] = term match {
       case Apply(TypeApply(Ident("as"), List(typeArg)), List(remainingCode)) => 
-        val toType = typeArg.tpe
         val fromType = remainingCode.tpe.widen
-        val action = FocusAction.CastAs(fromType, toType)
-        Some(Right(remainingCode, action))
+        val toType = typeArg.tpe
+
+        if (toType <:< fromType) {
+          val action = FocusAction.CastAs(fromType, toType)
+          Some(Right(remainingCode, action))
+        }
+        else Some(Left(FocusError.InvalidDowncast(fromType.show, toType.show)))
         
       case _ => None
     }
--- a/core/shared/src/main/scala-3.x/monocle/internal/focus/features/castas/CastAsParser.scala
+++ b/core/shared/src/main/scala-3.x/monocle/internal/focus/features/castas/CastAsParser.scala
@@ -1,0 +1,22 @@
+package monocle.internal.focus.features.castas
+
+import monocle.internal.focus.FocusBase
+
+private[focus] trait CastAsParser {
+  this: FocusBase => 
+
+  import macroContext.reflect._
+
+  object CastAs extends FocusParser {
+
+    def unapply(term: Term): Option[FocusResult[(Term, FocusAction)]] = term match {
+      case Apply(TypeApply(Ident("as"), List(typeArg)), List(remainingCode)) => 
+        val toType = typeArg.tpe
+        val fromType = remainingCode.tpe.widen
+        val action = FocusAction.CastAs(fromType, toType)
+        Some(Right(remainingCode, action))
+        
+      case _ => None
+    }
+  }
+}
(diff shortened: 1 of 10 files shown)


core/shared/src/test/scala-3.x/monocle/focus/FocusCastAsTest.scala
```diff
@@ -0,0 +1,54 @@
...
+    val mealAppleColor = Focus[Meal](_.mainIngredient.as[Apple].color)
+
+    val mealA = Meal(Apple(24, "green"))
+    val mealB = Meal(Banana(50, false))
+
+    assertEquals(mealAppleColor.getOption(mealA), Some("green"))
+    assertEquals(mealAppleColor.getOption(mealB), None)
+  }
+
+  test("Cast a narrow thing to a broad thing, which is useless but allowed") {
```
julien-truffaut, line 38: should we allow this or request the target to be a subtype of current target?
For example, `GenPrism` is currently defined with:
```scala
object GenPrism {
  def apply[S, A <: S]: Prism[S, A] = macro ...
}
```

julien-truffaut:
Good point, I think it makes sense to add `as` a method of all optics/apply optics. We can do that in a separate PR.

core/shared/src/main/scala-3.x/monocle/internal/focus/features/castas/CastAsParser.scala
```diff
@@ -0,0 +1,22 @@
...
+
+private[focus] trait CastAsParser {
+  this: FocusBase => 
+
+  import macroContext.reflect._
+
+  object CastAs extends FocusParser {
+
+    def unapply(term: Term): Option[FocusResult[(Term, FocusAction)]] = term match {
+      case Apply(TypeApply(Ident("as"), List(typeArg)), List(remainingCode)) =>
```
julien-truffaut, line 13: Should we require the current type to be a sealed trait / enum and `toType` to be a subtype of it?

julien-truffaut: This should be the only safe use case and the more constrained we are, the better the error message we can produce. WDYT?

core/shared/src/test/scala-3.x/monocle/focus/FocusCastAsTest.scala
kenbot: Good point! I'll put in this restriction and amend the PR.

core/shared/src/main/scala-3.x/monocle/internal/focus/features/castas/CastAsParser.scala
kenbot: I don't see any advantage at all in this restriction - what is less safe? I'm pretty sure a downcast will always be a valid Prism.

kenbot: Also, it wouldn't be supported in the standalone `as`, so it wouldn't commute with Focus

julien-truffaut: It is a more general discussion about what one should or should not be able to do. I would argue that we should only be able to pattern match/down cast on a sealed trait/abstract class or final class.
Let's say it otherwise, in which case would want to use `as` and the source type is not sealed?

This may not be the right place to have this discussion. I think it is awesome what you did with `as`, but in case you would like to echange more on this topic, I am happy to do it. Maybe not in this PR :)

julien-truffaut: > Also, it wouldn't be supported in the standalone `as`, so it wouldn't commute with Focus

Very good point. I guess we could do it if the standalone `as` was implemented via a macro (if that's possible), but even then, it wouldn't work for Scala 2.12. So you are right, let's not add any more constraints.

julien-truffaut (approved):

julien-truffaut:
Amazing work, thanks @kenbot

PR #1055, opened 2021-02-13
Having our cake and eating it

yilinwei:
Restricts the actions to happen only within the context. 

Hopefully IDE's are clever enough to figure this out.

**/cc** @kenbot

### DIFF
--- a/core/shared/src/main/scala-3.x/monocle/Focus.scala
+++ b/core/shared/src/main/scala-3.x/monocle/Focus.scala
@@ -3,16 +3,16 @@ package monocle
 import monocle.internal.focus.{FocusImpl, AppliedFocusImpl}
 import monocle.syntax.FocusSyntax
 
-object Focus extends FocusSyntax {
+object Focus {
 
   extension [From, To] (from: From) 
-    transparent inline def focus(inline lambda: (From => To)): Any = 
+    transparent inline def focus(inline lambda: (FocusSyntax ?=> From => To)): Any = 
       ${AppliedFocusImpl[From, To]('from, 'lambda)}
 
   def apply[S] = new MkFocus[S]
 
   class MkFocus[From] {
-    transparent inline def apply[To](inline lambda: (From => To)): Any = 
+    transparent inline def apply[To](inline lambda: (FocusSyntax ?=> From => To)): Any = 
       ${ FocusImpl('lambda) }
   }
 }
--- a/core/shared/src/main/scala-3.x/monocle/Focus.scala
+++ b/core/shared/src/main/scala-3.x/monocle/Focus.scala
@@ -1,23 +1,23 @@
 package monocle
 
-import monocle.internal.focus.{FocusImpl, AppliedFocusImpl}
+import monocle.internal.focus.{FocusImpl, AppliedFocusImpl, InFocus}
 
 object Focus {
 
-  extension [From, To] (from: From) 
-    transparent inline def focus(inline lambda: (From => To)): Any = 
+  extension [From, To] (from: From)
+    transparent inline def focus(inline lambda: (InFocus ?=> From => To)): Any =
       ${AppliedFocusImpl[From, To]('from, 'lambda)}
 
-  extension [CastTo] (from: Any)
+  extension [CastTo] (from: Any)(using InFocus)
     def as: CastTo = scala.sys.error("Extension method 'as[CastTo]' should only be used within the monocle.Focus macro.")
 
-  extension [A] (opt: Option[A])
+  extension [A] (opt: Option[A])(using InFocus)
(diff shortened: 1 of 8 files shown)


core/shared/src/main/scala/monocle/syntax/All.scala
```diff
@@ -2,4 +2,4 @@ package monocle.syntax
 
 object all extends Syntaxes
 
-trait Syntaxes extends ApplySyntax with FocusSyntax with FieldsSyntax
+trait Syntaxes extends ApplySyntax with FieldsSyntax
```
julien-truffaut, line 5: I just realised that the extension method for partially applying foucs was not in FocusSyntax. It is a mistake.

We need to add to test to show that users can do `user.focus(_.name).get` with `import monocle.syntax.all._`

core/shared/src/main/scala-3.x/monocle/Focus.scala
```diff
@@ -3,16 +3,16 @@ package monocle
 import monocle.internal.focus.{FocusImpl, AppliedFocusImpl}
 import monocle.syntax.FocusSyntax
 
-object Focus extends FocusSyntax {
+object Focus {
 
   extension [From, To] (from: From) 
-    transparent inline def focus(inline lambda: (From => To)): Any = 
+    transparent inline def focus(inline lambda: (FocusSyntax ?=> From => To)): Any =
```
kenbot, line 9: I like the name `FocusSyntax` a lot better than `InFocus`, which didn't carry as much meaning.

core/shared/src/main/scala-3.x/monocle/internal/focus/FocusImpl.scala
```diff
@@ -1,31 +1,37 @@
...
   }
 }
 
+
 private[monocle] object FocusImpl {
-  def apply[From: Type, To: Type](lambda: Expr[From => To])(using Quotes): Expr[Any] =
-    new FocusImpl(quotes).run(lambda)
-}
\ No newline at end of file
+  def apply[From: Type, To: Type](contextExpr: Expr[FocusSyntax ?=> From => To])(using Quotes): Expr[Any] = {
```
kenbot, line 31: There's something to be said for keeping the outermost layer completely free of logic. We can see already that being on the outside, it has to reinvent how errors are handled. The danger is that future-us will see this outer layer as a "place to put things", which it shouldn't be.

On the other hand, there probably won't be thousands of committers trying to put things there. As it currently is, it is already a clean compositional layer that neatly wraps and plugs into the existing functionality without changing it.

On balance, I don't really mind if this bit stays as it is.

core/shared/src/main/scala-3.x/monocle/internal/focus/InFocus.scala
```diff
@@ -0,0 +1,19 @@
+package monocle.internal.focus
+
+import monocle.syntax.FocusSyntax
+import scala.quoted.{Quotes, Type, Expr}
+import scala.annotation.implicitNotFound
+
+private[focus] object InFocus {
```
kenbot, line 7: There's no need for an "InFocus" concept now.  I think this method should be called `stripFocusContext` and moved to live inside  `FocusImpl` as a private method.

core/shared/src/main/scala-3.x/monocle/internal/focus/features/castas/CastAsParser.scala
```diff
@@ -3,24 +3,26 @@ package monocle.internal.focus.features.castas
...
-  this: FocusBase => 
+  this: FocusBase =>
 
   import macroContext.reflect._
 
   object CastAs extends FocusParser {
 
     def unapply(term: Term): Option[FocusResult[(Term, FocusAction)]] = term match {
-      case Apply(TypeApply(Ident("as"), List(typeArg)), List(remainingCode)) => 
+      case Apply(TypeApply(Select(_, "as"), List(typeArg)), List(remainingCode)) =>
```
kenbot, line 13: Generally, where some subset of these TASTy formulas correspond to some discrete purpose, we should extract that pattern out into a named extractor, like `CaseClass` or `FieldType` in the `SelectFieldParser` module. This makes the code more readable.

Here we could have, say, a `FocusKeyword` extractor (ie `Apply(TypeApply(FocusKeyword("as"), ...` that handles that extra Select.

core/shared/src/main/scala-3.x/monocle/syntax/FocusSyntax.scala
```diff
@@ -1,6 +1,6 @@
 package monocle.syntax
 
-trait FocusSyntax {
+sealed trait FocusSyntax {
```
kenbot, line 3: As above, let's pack this up into `Focus.scala`

PR #1058, opened 2021-02-13
add Focus to all syntax import

julien-truffaut:
This PR adds all the `Focus` extensions `as`, `some`, etc to the import `monocle.syntax.all._`

I wonder if we should create a simpler import such as `monocle.all._` or `monocle.implicits._`

### DIFF
--- a/core/shared/src/main/scala-2.x/monocle/syntax/FocusSyntax.scala
+++ b/core/shared/src/main/scala-2.x/monocle/syntax/FocusSyntax.scala
@@ -1,0 +1,3 @@
+package monocle.syntax
+
+trait FocusSyntax
--- a/core/shared/src/main/scala-3.x/monocle/Focus.scala
+++ b/core/shared/src/main/scala-3.x/monocle/Focus.scala
@@ -1,19 +1,14 @@
 package monocle
 
 import monocle.internal.focus.{FocusImpl, AppliedFocusImpl}
+import monocle.syntax.FocusSyntax
 
-object Focus {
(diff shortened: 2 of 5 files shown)


julien-truffaut:
Conflict with https://github.com/optics-dev/Monocle/pull/1055

yilinwei:
IMO, I think that `.implicits` don't make any more sense for Dotty and I think that we provide any instances that can't live under the associated companion object.

Possibly a little controversial but maybe something like `monocle.dsl._` or `monocle.ext._` would capture intent better?

kenbot (approved):
I think it makes sense to integrate Focus syntax into the "syntax all" import thing, if someone is coming from that point of view. A more memorable import syntax for it might be better if this is something that we expect users to do a lot - import all of the syntax at once, that is.

It needn't concern us for this PR, but something to think about: given the "front door" position we are going to give Focus in the doco, we need to decide what imports to recommend as a default, and stick to it in examples. For instance, `import monocle.{_, Focus._}`

julien-truffaut:
@kenbot good point, I think we will probably use `import monocle._` to get `Focus` and all the optics in scope.

Then we need a second one, easy to remember for all extension methods within Monocle. Ideally, short, memorable and consitent with other API.

I quite like `import monocle.dsl._` but I am not sure if it is "standard".

yilinwei:
@julien-truffaut 

I know [http4s](https://http4s.org/v0.18/dsl/) do it for their imports which is where I originally got the idea from.

PR #1059, opened 2021-02-13
Simpler landing page with Focus

julien-truffaut:
Fix https://github.com/optics-dev/Monocle/issues/1057

### DIFF
--- a/website/i18n/en.json
+++ b/website/i18n/en.json
@@ -3,7 +3,7 @@
   "localized-strings": {
     "next": "Next",
     "previous": "Previous",
-    "tagline": "Optics for Scala",
+    "tagline": "Access and transform immutable data",
     "docs": {
       "examples": {
         "title": "Examples"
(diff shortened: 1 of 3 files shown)


yilinwei:
LGTM.

It would be cool if we could embed a console so users could try out the common snippets. Might try out a prototype to see if it feels good.

julien-truffaut:
Maybe we can integrate it with scatsie?

yilinwei:
Yes - it looks doable. See [here](https://codepen.io/MasseGuillaume/pen/oeKoee). [1]

[1] Taken from the scastie README

kenbot (approved):
lgtm!

julien-truffaut:
> Yes - it looks doable. See [here](https://codepen.io/MasseGuillaume/pen/oeKoee). [1]
> 
> [1] Taken from the scastie README

Maybe we could have a section "Play with Monocle" that would be a set of exercises with scatsie?

PR #1060, opened 2021-02-14
Focus[Foo]() as entry point to Monocle

julien-truffaut:


kenbot (commented):
I like the idea of being able to do an "id" Focus, it certainly makes the commuting laws easier to express. 

Rather than having a special no-arg form, I think it would be more consistent to keep the lambda function, and just allow the user to provide an identity function. eg `Focus[Foo](_)`, or in other words `Focus[Foo](a => a)`.

Should we therefore provide `Focus.id[A]` by analogy to `Iso.id`? It is strictly redundant of course, but if Focus is the main conduit for users' optics needs, it might come to mind easier for them.

julien-truffaut:
IMO, `id` makes sense when you talk about  specific`Optics` because there is a relationship between `id` and `andThen`. However, `Focus` is more a path builder, in that context I think `Focus[User]()` or `Focus[User](_)` (as you proposed) represents the "empty" path.

Another option could `Focus.root[User]`. 

I don't have a strong opinion either way, I am just trying to guess what would feel the more natural for a beginner.

yilinwei:
Sleeping on it, I think I prefer `Focus(_)` or `Focus(x => x)`. It allows us to keep the DSL more consistent - especially when we switch to creating an `Iso` for a single field case class.

julien-truffaut:
I am fine with `Focus[User](_)` but I am not surewhat should be the signature? Would it require special treatment in the macro?

julien-truffaut:
I created https://github.com/optics-dev/Monocle/issues/1064 to track it

PR #1062, opened 2021-02-15
Try new naming convention on Lens

julien-truffaut:
This is a poc to see how the naming conventions we discussed would look like in practice.

### DIFF
--- a/core/shared/src/main/scala/monocle/Lens.scala
+++ b/core/shared/src/main/scala/monocle/Lens.scala
@@ -29,13 +29,13 @@ abstract class PLens[SourceIn, SourceOut, TargetIn, TargetOut] extends Serializa
   /** modify the target using a function with effects */
   def modifyF[Effect[_]: Functor](f: TargetIn => Effect[TargetOut])(s: SourceIn): Effect[SourceOut]
 
-  /** modify the target using a function. 
+  /** modify the target using a function.
     * This is equivalent to: {{
     *   val target    = get(source)
     *   val newTarget = update(target)
     *   replace(newTarget)(source)
     * }}
-    * */
+    */
   def modify(update: TargetIn => TargetOut): SourceIn => SourceOut
 
   /** find if the target satisfies a predicate */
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/monocle/Lens.scala
```diff
@@ -5,183 +5,207 @@ import cats.arrow.Choice
...
       (self.get(s), other.get(s1))
     } {
       case (b, b1) => { case (s, s1) =>
         (self.replace(b)(s), other.replace(b1)(s1))
       }
     }
 
-  final def first[C]: PLens[(S, C), (T, C), (A, C), (B, C)] =
-    PLens[(S, C), (T, C), (A, C), (B, C)] { case (s, c) =>
+  final def first[C]: PLens[(SourceIn, C), (SourceOut, C), (TargetIn, C), (TargetOut, C)] =
```
julien-truffaut, line 69: I am not sure what to do here.

kenbot: Just plain old `A` I'd say

julien-truffaut:
> Some of the TargetIns are mixed up with TargetOuts

Do you have an example? Sorry I am starting to mix things out ...

julien-truffaut:
ah I thought it was a typo from in your original suggestion. I don't get why `A` is an OUT and `B` is an IN.
It seems that IN implies covariance and OUT contravariance?

julien-truffaut:
ahah I can never get variance correctly. What I really struggle with is:
`TargetOut = target in the normal output position = A`
Sorry for being slow.

joroKr21:
> What I struggle with is why modify is (TargetOut => TargetIn) => SourceIn => SourceOut.

Where do you see that?

I'm not sure source and target give the right intuition though. We use optics that give us a view of some data type (whether it's zooming in with a lens, refracting with a prism, etc). How about something like `ViewIn => ViewOut` and `DataIn => DataOut`?

julien-truffaut:
> Variance flips around in function arguments. A function type on its own is `- => +`, an argument is `-`, so a function type argument is `+ => -`. 
> 
> 
> 
> Basically no devs understand this though, so it's a crappy intuition to rely on

Yeah I understand the variance of each parameter though in my mind the old A is not an Out.

julien-truffaut:
Closing, we need more discussion.

Issue #1029, opened 2021-02-08
[Focus] zoom into all the elements of a collection using each


julien-truffaut:
I think this differs from the idx and ? features in that it might result in a different kind of optic, and EPTraversal. Currently the code casts the whole macro to Lens[From,To] at the end, but now we don't know what the return is necessarily going to be.

We could use the new Scala 3 transparent keyword for "white box macro" methods,

PR #1072, opened 2021-02-17
[Focus] Each feature

kenbot:
Adds `each` to the Focus macro, resolving #1029.

Some more bits and pieces:
- Fixed the ridiculous `composeOptics` method to be two lines, owing to a horrendous Dotty compiler space leak the old version was causing now that Traversals are in the mix.
- Removed the `ParseResult` type alias. Give it to me straight
- Removed the fully qualified `_root_.monocle.this.that....`, which is a habit I had picked up writing macros in stringier environments. They are real typechecked Scala symbols, there's no chance of ambiguity or accidental capture of some other symbol by the same name.

### DIFF
--- a/core/shared/src/main/scala-3.x/monocle/Focus.scala
+++ b/core/shared/src/main/scala-3.x/monocle/Focus.scala
@@ -1,5 +1,6 @@
 package monocle
 
+import monocle.function.Each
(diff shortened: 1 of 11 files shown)


julien-truffaut:
It looks great!

julien-truffaut (approved):
Great job

PR #1073, opened 2021-02-17
Start focus documentation, add Focus to Scala 2 which redirect to GenLens

julien-truffaut:
The idea is to make the Scala 2 and Scala 3 API as similar as possible to simplify the migration and documentation.
`Focus` will be weaker in Scala 2, only lenses for fields and it will require a different import `monocle.macro.syntax.all._`

- Add `Focus` in Scala 2 macro module which redirect to `GenLens`
- Add `focus` extension method in Scala 2 which creates an `ApplyLens`
- Start documenting `Focus` i nthe website for both Scala 3 and Scala 2. There is no version of mdoc for Scala 3 yet, so the scala 3 examples are not compiled

### DIFF
--- a/docs/src/main/mdoc/focus.md
+++ b/docs/src/main/mdoc/focus.md
@@ -1,0 +1,105 @@
+---
+id: focus
+title: Focus
+---
+
+`Focus` is the best starting point into Monocle. `Focus` lets you define a path within an immutable object. 
+Then, once you reach your desired target, you can just as easily get, replace or modify it. Let’s have a look at the 
+most common use cases.
+
+An important point before we start. `Focus` is a macro available for both Scala 2.13 and Scala 3. 
+However, the macro API has completely changed between Scala 2 and 3, so for each example, we will first show the 
+version for Scala 3 and then Scala 2 (often more verbose). 
+
+## Update a field of a case class
+
+```scala mdoc:silent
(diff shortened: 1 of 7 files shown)


docs/src/main/mdoc/focus.md
```diff
@@ -0,0 +1,105 @@
+---
+id: focus
+title: Focus
+---
+
+`Focus` is the best starting point into Monocle. `Focus` lets you define a path within an immutable object.
```
kenbot, line 6: "...within a graph of immutable objects..."?

docs/src/main/mdoc/focus.md
```diff
@@ -0,0 +1,105 @@
...
+anna
+  .focus(_.address.streetNumber)
+  .modify(_ + 1)
+// res: User = User(
+//   name = "Anna",
+//   address = Address(streetNumber = 13, streetName = "high street")
+// )
+```
+
+The Scala 2 version is the same, except for the import
```
kenbot, line 44: Agree with you that Monocle 2 and 3 instructions should ideally be in different pages

kenbot (approved):
Doco reads well to me, other than the clutter of having Monocle 2 & 3 stuff side by side. Smart move rebadging `.lens` and friends. Smooths the path to Monocle 3.

PR #1079, opened 2021-02-18
Yilin's powerful syntax technique

kenbot:
Introduces Yilin's powerful new syntax technique, replacing PR #1055. 

- Starts from Yilin's commits, preserving it in the history
- Has cake and also eats it, 🎂 nom nom nom
- Moves the Focus keywords from `monocle.syntax.FocusSyntax` into a `monocle.Focus.KeywordContext` trait, because it not like the other `xxxSyntax` traits and shouldn't live with them or look like them. Alternative naming suggestions welcome
- Moves the `extension ... def focus` pre-applied Focus into an `AppliedFocusSyntax`, which is like the other syntax traits
- ~Haven't mixed that into `monocle.syntax.all` because I haven't figured out how to do that correctly with the different Scala version directories, suggestions welcome~ I figured it out. Lmk if you want me to remove it from `object Focus` as well
- Fixed up error messages that still somehow saying "GenLens"
- Moved the keyword stripping logic inside the macro, because I considered that it belongs in there, it is not truly an external bolt-on. You can't separate the internal logic from the fact that the keyword context was passed in
- Split the general lambda function configuring out from `ParserLoop` into a new thing `LambdaConfigParser`, which also does the keyword magic stripping. These boundaries make sense, because this initial config stuff is very slow moving and should not change often. The feature loop stuff in `ParserLoop` however, is fast moving and often needs to change.
- `ParserParams` renamed to `ParserConfig`, and unused type argument removed.


yilinwei (approved):

yilinwei:
LGTM.

Raised #1080 to track the rest of the issue.

simlei:
Hi, I'm a total outsider here; but I got very interested seeing the `KeywordContext ?=>` stuff in Monocle3. I tried to figure it out by reading https://github.com/optics-dev/Monocle/pull/1079 — but I think an insider's explanation would be super helpful. Would someone care to elaborate or link me to elaborations of how cake is had and eaten too, and how this is probably a Scala 3 feature to be desired by generic library authors? I shall be most happy :)

simlei:
Hi @kenbot, thank you very much for taking the time to outline this. I think this clears it up for me :) Also, this is the example most useful for me for learning about context lambdas that I have seen to date.
Cheers :)

julien-truffaut:
@kenbot I wonder if there would be a benefit to mark `KeywordContext` as `erased` https://docs.scala-lang.org/scala3/reference/experimental/erased-defs.html

PR #1081, opened 2021-02-20
'At', 'index' features + refactor of parsers for readability and type safety

kenbot:
Adds `at` and `index` features, resolving #1030.

Of note:
- `at` feature preserves singleton index types, so as to support tuple access in the same way as `monocle.function.At`.
- Big refactor of the parser matchers for readability and type safety, based on pulling out the obvious commonality between the matchers
- All of the features with keywords (ie not field select) now use the new pretty extractors with descriptive case classes wrapping outputs, and no longer need access to the `macroContext.reflect` object.
- As usual, the tests are fairly light, and will need to be made more comprehensive later
- New module `ParserBase` that contains the `FocusParser` interface, the new `FocusKeyword` and `FocusKeywordGiven` matches and all the typesafe case class wrappers.
- `ParserBase`, `ParserLoop` and `GeneratorLoop` now live in `monocle.internal.focus.features`, closer to the features they support and aggregate.

### DIFF
--- a/core/shared/src/main/scala-3.x/monocle/internal/focus/features/castas/CastAsParser.scala
+++ b/core/shared/src/main/scala-3.x/monocle/internal/focus/features/castas/CastAsParser.scala
@@ -1,21 +1,16 @@
 package monocle.internal.focus.features.castas
 
 import monocle.internal.focus.FocusBase
+import monocle.internal.focus.features.ParserBase
 
 private[focus] trait CastAsParser {
-  this: FocusBase =>
-
-  import macroContext.reflect._
+  this: FocusBase with ParserBase =>
 
   object CastAs extends FocusParser {
 
-    def unapply(term: Term): Option[FocusResult[(Term, FocusAction)]] = term match {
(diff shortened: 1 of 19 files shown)


core/shared/src/main/scala-3.x/monocle/internal/focus/features/castas/CastAsParser.scala
```diff
@@ -1,21 +1,16 @@
...
-    def unapply(term: Term): Option[FocusResult[(Term, FocusAction)]] = term match {
-      case Apply(TypeApply(FocusKeyword("as"), List(typeArg)), List(remainingCode)) =>
-
-        val fromType = remainingCode.tpe.widen
-        val toType = typeArg.tpe
-
-
+    def unapply(term: Term): Option[FocusResult[(RemainingCode, FocusAction)]] = term match {
+      
+      case FocusKeyword(Name("as"), FromType(fromType), TypeArgs(toType), remainingCode) =>
```
julien-truffaut, line 13: it reads very well, great job

julien-truffaut:
Great job! if you have time, could you also update the example in `focus.md`?

julien-truffaut (approved):

Issue #996, opened 2020-12-18
Use inheritance between optics


julien-truffaut:
Currently, Optics have a few methods `asX`, e.g. `asTraverse`, `asOptional` on `Lens` to "upcast" optics.

Instead, we could introduce inheritance between optics, e.g. 

```scala
trait Lens[S, T, A, B] extends Optional[S, T, A, B] with Getter[S, A]
```

Then we can deprecate all the `asX` methods.

julien-truffaut:
There is an issue with overriding `modifyF`

https://scastie.scala-lang.org/XVL7KRPEQo6wfRy3GgtUpw

jdegoes:
Please do! 🙏

kenbot:
I think inheritance "tells the truth" about the relationship between the optics. As long as it doesn't do anything crazy for type inference or variance or whatever, I think it's a good idea. 

You shouldn't weigh this heavily at all, I signed up for difficulty, but it will certainly make Monocle 3 macro-writing slightly more convenient.

julien-truffaut:
I made some progress but unfortunately it doesn't look good.

As @tpolecat told me in gitter, Scala doesn't accept narrowing on inputs https://gitter.im/scala/scala?at=602fd83b9ba13e72e43d7ea1

We could get around that by defining two version of `modifyF` for `Lens`, one that takes a `Functor` and one that takes an `Applicative` and write one in terms of the other. This trick works if all the arguments are within the same parentheses but it doesn't work if they are split over several parameterer lists :(

https://scastie.scala-lang.org/GPj1KhiBQnSsoXCACTLEVw

julien-truffaut:
Since there is only one method affected:  `modifyF`, maybe we could consider tto remove the overload and do:
```scala
trait Lens[S, T, A, B] {
  def foo[F[_]: Functor    ](update: A => F[B])(from: S): F[T]
  def bar[F[_]: Applicative](update: A => F[B])(from: S): F[T]
}
```

but I am not sure how to call `foo` and `bar`

joroKr21:
`modifyF -> Functor` and `modifyA -> Applicative`

julien-truffaut:
Inheritance is overloading some methods because they have slightly different shape for optics with 2 or 4 type parameters, see https://github.com/optics-dev/Monocle/pull/1082.

julien-truffaut:
Implemented in #1088

PR #1082, opened 2021-02-20
WIP use inheritance between optics and variance

julien-truffaut:
Couple of points to note:

- I haven't touched `ApplyOptics` in this PR, I propose to do it separately
- I added `modifyA` instead of `modifyF` when `F` is an `Applicative`, not great but iot works around the narrowing issue mentioned here #996
- I moved shortcuts `each`, `at`, ... from `Getter` and `Fold` to an implicit class to avoid those methods taking precendence
- deprecated `Optic.id` and `Optic.void` except for `Iso.id` and `Optional.void`

One test doesn't compile, I am not sure why.


julien-truffaut:
Inheritance is making some methods overloaded because Read-Only optics (`Getter` and `Fold`) have different methods shape.

```
[error] /Monocle/core/shared/src/test/scala/monocle/CompositionGenericOpticsTest.scala:13:10: overloaded method composePrism with alternatives:
[error]   [C, D](other: monocle.PPrism[Either[String,Int],Either[String,Int],C,D]): monocle.PPrism[Either[String,Int],Either[String,Int],C,D] <and>
[error]   [C, D](other: monocle.PPrism[Either[String,Int],Nothing,C,D]): monocle.Fold[Either[String,Int],C]
[error]  cannot be applied to (monocle.Prism[Either[Nothing,Nothing],Nothing])
[error]     root.composePrism(stdRight)
[error]          ^
```

Previously, `composePrism` had different number of type parameters 3 in `Getter` and `Fold`,  2 for the others. I tried to add variance so that we can use `Nothing` instead of a free type variable, but it still creates an overloaded method.

julien-truffaut:
I printed the overloaded methods of `Iso`:
-  "composePrism" -> 5,
-  "left" -> 4,
-  "to" -> 2,
-  "second" -> 5,
-  "split" -> 3,
-  "first" -> 5,
-  "choice" -> 6,
-  "composeLens" -> 6,
-  "composeGetter" -> 2,
-  "some" -> 5,
-  "right" -> 4,
-  "composeOptional" -> 4,
-  "adapt" -> 8,
-  "composeIso" -> 8,
-  "composeTraversal" -> 3

julien-truffaut:
I will reopen one with more documentation.

PR #1083, opened 2021-02-20
add test suite to check when composition works or not

julien-truffaut:
`CompositionGenericOpticsTest` make me think we should still offer non-overloaded compose methods.

We could add `andThenLens`, `andThenOptional`, ... but recommend to use them only when facing type inference issue which should hopefully be rare.

### DIFF
--- a/core/shared/src/test/scala/monocle/CompositionConcreteOpticsTest.scala
+++ b/core/shared/src/test/scala/monocle/CompositionConcreteOpticsTest.scala
@@ -1,0 +1,330 @@
+package monocle
+
+final class CompositionConcreteOpticsTest extends munit.FunSuite {
+
+  val iso: Iso[Int, Int] = Iso.id
+  val lens: Lens[Int, Int] = Lens.id
+  val prism: Prism[Int, Int] = Prism.id
+  val optional: Optional[Int, Int] = Optional.id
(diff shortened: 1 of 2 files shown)


yilinwei:
This will break once inheritance is added right? (the subclass would be valid).

julien-truffaut:
@yilinwei What would break? I think we should postpone the decision on keeping non-overloaded method once we see the impact of inheritance + variance in Scala 2 and 3.

yilinwei:
Sorry, "breaks" is the wrong word. Let me be a bit more specific; if the return type of the composition is a sub-type the test still works. I wonder if this form is better so it doesn't pass even if the sub-type is returned.

```scala
sealed trait Foo

trait Bar extends Foo
trait Baz extends Bar

object Bar {
  def apply(): Bar = ???
}

sealed trait ¬[A]

sealed trait NotLowPriority {
  given notambig[A]: ¬[A]  = ???
}

object ¬ extends NotLowPriority {
  given ambig1 [A](using a: A): ¬[A] = ???
  given ambig2 [A](using a: A): ¬[A] = ???
}


object Test {
  val optic = Bar()
  summon[optic.type <:< Bar]
  summon[¬[optic.type <:< Baz]]
}
```

julien-truffaut:
I added thoses tests directly to #1088

PR #1084, opened 2021-02-21
Added test case to confirm that private case class fields can be accessed by Focus

kenbot:
The problem raised in #498, that the `GenLens` macro did not allow the creation of Lenses on private case class fields, works fine with `Focus` in Monocle 3. 

I have added a test case to lock in this behaviour.

### DIFF
--- a/core/shared/src/test/scala-3.x/monocle/focus/FocusFieldSelectTest.scala
+++ b/core/shared/src/test/scala-3.x/monocle/focus/FocusFieldSelectTest.scala
@@ -1,6 +1,6 @@
 package monocle.focus
 
-import monocle.Focus
+import monocle.{Focus, Lens}
 
 final class FocusFieldSelectTest extends munit.FunSuite {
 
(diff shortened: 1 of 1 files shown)


core/shared/src/test/scala-3.x/monocle/focus/FocusFieldSelectTest.scala
```diff
@@ -20,6 +20,11 @@ final class FocusFieldSelectTest extends munit.FunSuite {
   case class ConstraintBox[A <: AnyVal](a: A)
   case class Varargs[A](a: A*)
 
+  object Public {
+    case class Private(private[Public] a: Int)
+    def lens: Lens[Private, Int] = Focus[Private](_.a)
```
julien-truffaut, line 25: out of curiosity, what happens if you try to create this lens outside of `Public`?

julien-truffaut (approved):
LGTM

PR #1085, opened 2021-02-21
More consistent naming for Focus features

kenbot:
The inconsistent naming of Focus features was annoying me. 

The package names are now all exactly the keyword, and the `FocusAction` and matcher/parser are now called `KeywordXXX` to avoid naming collisions on things like `Some`, but still be consistent for all of them.

`FieldSelect` stays how it is. It will probably end up splitting into features called something like `SelectLens`, `SelectIso`, `SelectMulti`, `SelectGetter`, etc, so it will need its own pattern.


cquiroz (approved):
LGTM

PR #1086, opened 2021-02-21
Add inheritance and variance between optics

julien-truffaut:
This is a massive PR, there might be a way to break it down but many problems are interconnected.

## Inheritance between optics `composeXXX`

Inheritance introduced regressions in type inference. This is because inheritance overloaded some methods. For example, the code below compiles without inheritance but fails afterward:

```scala
IIso.id[Either[String, Int]].composePrism(stdRight)
```

The compiler says that their are two overloaded versions of `composePrism`. This makes sense if you look at the signature of `Getter` and `PLens`.

```scala
trait Fold[S, A] {
    def composePrism[B, C, D](other: PPrism[A, B, C, D]): Fold[S, C] = ???
}

trait POptional[S, T, A, B] extends PTraversal[S, T, A, B] with Fold[S, A] {
  override def composePrism[C, D](other: PPrism[A, B, C, D]): POptional[S, T, C, D] = ???
}
```
Note that `composePrism` for `Fold` has 3 type parameters while the same method for `POptional` has 2 type parameters.

I thought that variance could help:
```scala
trait Fold[-S, +A] {
    def composePrism[C, D](other: PPrism[A, Nothing, C, D]): Fold[S, C] = ???
}

trait POptional[-S, +T, +A, -B] extends PTraversal[S, T, A, B] with Fold[S, A] {
  override def composePrism[C, D](other: PPrism[A, B, C, D]): POptional[S, T, C, D] = ???
}
```

Unfortunately, it still overlaod `composePrism`.

So I moved all the `composeXXX` and symbolic operator to extension method, this way they don't overlaod anything.

## `modifyF`

I found a limitation of Scala with regards to inheritance. As @tpolecat [mentioned](https://gitter.im/scala/scala?at=602fd83b9ba13e72e43d7ea1) it, Scala doesn't accept narrowing on inputs. This means we can't do:

```scala
trait Traversal[-S, +T, +A, -B] {
  def modifyF[F[_]](update: A => B)(s: S)(implicit ev: Applicative[F]): F[T @uncheckedVariance]
}

trait Lens[-S, +T, +A, -B]  extends Traversal[S, T, A, B]{
  // Overrides nothing, even though Functor is an Applicative
  override def modifyF[F[_]](update: A => B)(s: S)(implicit ev: Functor[F]): F[T @uncheckedVariance]
}
```

The only solution I could come up with is to rename `modifyF` in `Traversal`, `Optional` and `Prism` to `modifyA` and keep `modifyF` for `Lens` and `Iso`.  If somone finds a better name, we can rename it later. Thanks to @joroKr21 for the suggestion.

## `some`

Scala 2 type inference doesn't work with the new signature of `some` method, but Scala 3 is fine:
```scala
def some[A1, B1](implicit ev1: A <:< Option[A1], ev2: Option[B1] <:< B): PTraversal[S, T, A1, B1] = ???
```

One solution is to limit `some` to monomorphic cases and move it to an extension method  in `XXXSyntax`. The tests pass but then the cats extension method `some` takes precedence (which broke the website).

So I am not sure what to do with this case ...

## uncheckedVariance

I sprinkled the code with `uncheckedVariance`, notably for `modifyF` because cats `Functor`/`Applicative` are invariant. I have no idea if it is safe or not.

## `adapt` with `asInstanceOf`

I implemented `adapt` with `asInstanceOf` because Scala 2 required because Scala 2 required variance annotation in `substituteContra` and `substituteCo` but Scala 3 didn't accept them. See https://scastie.scala-lang.org/yN70A1MHRgGEFpB6cyzmtQ

I will to try to rollback variance and keep the other changes to see if it solve the issue with `some`.


joroKr21:
I forgot to mention that another option to choosing another name is using `DummyImplicit` to make the overloading work. Neither is ideal of course.

julien-truffaut:
@joroKr21 could you show me how would that work?

julien-truffaut:
Closing in favor of #1088 which only introduce inheritance

PR #1093, opened 2021-02-23
use slash syntax instead of old "in" syntax

xuwei-k:
https://github.com/sbt/sbt/pull/6328


julien-truffaut (approved):
LGTM, thanks @xuwei-k

Issue #1031, opened 2021-02-08
[Focus] add `withDefault` feature


julien-truffaut:
`withDefault(...)` allows to fall back to a value in case the Option is a None.

```scala
case class User(name: String, address: Option[Address])
case class Address(streetNumber: Int, postcode: String)

val elise = User("Elise", Some(Address(12, "high street")))
val bob   = User("bob"  , None)

elise.focus(_.address.some.streetNumber).replace(24)
// res: User = User("Elise", Some(Address(24, "high street")))

// no-op address is None
bob.focus(_.address.some.streetNumber).replace(24) 
// res: User = User("bob", None) 

// same as some
elise.focus(_.address.witdDefault(Address(1, "")).streetNumber).replace(24)
// res: User = User("Elise", Some(Address(24, "high street")))

// same as some
bob.focus(_.address.witdDefault(Address(1, "")).streetNumber).replace(24)
// res: User = User("bob", Some(Address(24, ""))) 
```

`withDefault` generates an `Iso` which is not completely correct see example here https://github.com/optics-dev/Monocle/blob/a6506a3623455936947220ef15bb071c478727e3/core/shared/src/main/scala/monocle/std/Option.scala#L25-L41

kenbot:
Given our previous conversation on this, I think `withDefault` is perfectly fine and we should stick with it

PR #1094, opened 2021-02-23
Focus withDefault feature, removed Eq constraint on withDefault everywhere

kenbot:
Focus `withDefault` feature, resolving #1031 

Of note: 
- Removed the `Eq` constraint from `withDefault` in `std.option` and in all the optics, because it adds an intolerable usability hurdle to Focus, and provides little value for the `withDefault` usecase on its own
- Tweaked the `FocusKeyword` matcher in `ParserBase` to accomodate value parameters with no typeclass

### DIFF
--- a/core/shared/src/main/scala-3.x/monocle/internal/focus/features/ParserLoop.scala
+++ b/core/shared/src/main/scala-3.x/monocle/internal/focus/features/ParserLoop.scala
@@ -8,6 +8,7 @@ import monocle.internal.focus.features.as.AsParser
 import monocle.internal.focus.features.each.EachParser
 import monocle.internal.focus.features.at.AtParser
 import monocle.internal.focus.features.index.IndexParser
+import monocle.internal.focus.features.withdefault.WithDefaultParser
 
 private[focus] trait AllFeatureParsers 
   extends FocusBase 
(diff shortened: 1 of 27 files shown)


core/shared/src/main/scala-3.x/monocle/internal/focus/features/ParserLoop.scala
```diff
@@ -46,6 +48,9 @@ private[focus] trait ParserLoop {
         case KeywordAs(Right(remainingCode, action)) => loop(remainingCode, action :: listSoFar)
         case KeywordAs(Left(error)) => Left(error)
 
+        case KeywordWithDefault(Right(remainingCode, action)) => loop(remainingCode, action :: listSoFar)
```
julien-truffaut, line 51: I wonder if there is a way to capture this pattern of `KeyWord(Left(e)) => e`

julien-truffaut (approved):
LGTM, great work Ken!

core/shared/src/main/scala-3.x/monocle/internal/focus/features/ParserLoop.scala
kenbot: No doubt, there's a few drops of juice left to squeeze out. It's not a priority for me at this point, it's completely sterile as far as unwanted code growth is concerned, and squeezing harder risks overfitting.

PR #1095, opened 2021-02-23
Focus now supports tuple field selection

### LINKED ISSUE #1092: [Focus] Direct support for tuples fields _1, _2, etc
We should support `_1`, `_2` etc inside the Focus macro for tuples automatically. Currently this doesn't work

kenbot:
Resolves #1092. Also added some test cases.


julien-truffaut (approved):
LGTM!

PR #1097, opened 2021-02-23
Slightly improve university example

julien-truffaut:
They were some errors in this MD, non required import and use of `GenLens`.

Maybe better to ditch this file completely.


kenbot (approved):
lgtm

PR #1098, opened 2021-02-24
add Focus[X]() and .focus() for Iso/ApplyIso

### LINKED ISSUE #1064: Focus[X](_) as beginner-friendly version of `Iso.id[X]`
The idea is to delay the moments users need to learn about specific optics type.

julien-truffaut:
Fix #1064

## Use paren

I could make `.focus` work but not `Focus[X]`. The problem is that `Focus` uses the trick to partially apply type parameter
```scala
def apply[S] = new MkFocus[S]

class MkFocus[From] {
    def apply: Iso[From, From] = Iso.id
}
```

as a result `Focus[X]` is a `MkFocus[X]` not an `Iso[X, X]`. Adding paren to `apply` fixed the ambiguity.

## Moved syntax

I had to move `.focus()` next to the macro extension `.focus(_.foo.bar)`. When both extension methods where in different classes, it confuses the compiler. 

The main issue is for Scala 2, we need define `.focus()` in the macro module.

### DIFF
--- a/core/shared/src/main/scala/monocle/syntax/Apply.scala
+++ b/core/shared/src/main/scala/monocle/syntax/Apply.scala
@@ -5,8 +5,6 @@ import monocle._
 object apply extends ApplySyntax
 
 trait ApplySyntax {
-  implicit def toApplyOpticsOps[S](value: S): ApplyOpticsOps[S] = ApplyOpticsOps(value)
-
   implicit def toApplyFoldOps[S](value: S): ApplyFoldOps[S] =
     new ApplyFoldOps(value)
   implicit def toApplyGetterOps[S](value: S): ApplyGetterOps[S] =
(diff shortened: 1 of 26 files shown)


core/shared/src/main/scala/monocle/syntax/Apply.scala
```diff
@@ -25,79 +23,75 @@ trait ApplySyntax {
...
 final case class ApplyFoldOps[S](private val s: S) extends AnyVal {
-  @deprecated("use optics.andThen", since = "3.0.0-M1")
+  @deprecated("use focus.andThen", since = "3.0.0-M1")
   def applyFold[A](fold: Fold[S, A]): ApplyFold[S, A] =
     new ApplyFold[S, A](s, fold)
 }
 
 final case class ApplyGetterOps[S](private val s: S) extends AnyVal {
-  @deprecated("use optics.andThen", since = "3.0.0-M1")
+  @deprecated("use focus.andThen", since = "3.0.0-M1")
```
kenbot, line 33: Doco should reflect paren usage `focus()`

kenbot (changes requested):
Just do the doco fix and you're good to go

core/shared/src/main/scala/monocle/syntax/Apply.scala
julien-truffaut: good catch

kenbot (approved):
Lgtm

PR #1100, opened 2021-02-24
generalise Traversal.applyN to Optional

julien-truffaut:


core/shared/src/main/scala/monocle/Traversal.scala
```diff
@@ -185,14 +186,18 @@ object Traversal {
...
     PTraversal.apply6(get1, get2, get3, get4, get5, get6)(set)
 
-  /** Composes N lenses horizontally.  Note that although it is possible to pass two or more lenses
-    * that point to the same `A`, in practice it considered an unsafe usage (see https://github.com/julien-truffaut/Monocle/issues/379#issuecomment-236374838).
+  /** Merge multiple Optionals together.
+    * All Optional must target different piece of data otherwise the Traversal doesn't respect all properties.
+    * See this thread for more details: https://github.com/julien-truffaut/Monocle/issues/379#issuecomment-236374838.
     */
-  def applyN[S, A](xs: Lens[S, A]*): Traversal[S, A] =
+  def applyN[S, A](xs: Optional[S, A]*): Traversal[S, A] =
```
cquiroz, line 193: So we won't have `applyN` for lenses anymore?

julien-truffaut: No, it still work on Lenses because we introduce inheritance between optics, so a `Lens` is an `Optional`

cquiroz: Cool, I missed that

cquiroz (approved):
LGTM

Issue #1087, opened 2021-02-22
Question: Would you consider implementing some cats instances for optics?


vaslabs:
Hello,
I understand that my use case maybe a bit peculiar but anyway, long story short I needed to do something like 
[this](https://github.com/vaslabs/Monocle/commit/5216d912ce415a81f230ff701b08c6339738f5c4#diff-007ad103a52ea4afe202ac763485c26b2e5a3bf397696e5af026676455d08ff3R73)

Which needed `Invariant[Lens]` and `Semigroupal[Lens]` (or InvariantSemigroupal but I kept them separate in this linked snippet)

Would you like this kind of thing (i.e. implementing some cats instances in monocle) here? Could be a separate module dedicated for cats (I've seen there's State in a separate module), although that could mean a separate import to make `cats.syntax._` work

Let me know your thoughts, I could contribute the ones in the snippet (and possibly more?).

Edit:
I've also haven't looked into doing this for the full `PLens[S, T, A, B]`

julien-truffaut:
Hey @vaslabs and thanks for the question.

We were planning to add similar functionalities https://github.com/optics-dev/Monocle/issues/1019

The benfit of using cats `Semigroupal` is that we get all the syntax out of the box but we need an extra import for cats syntax. I don't have a strong opinion about it.

We already have cats instances within the `core` module, e.g. `Choice[Lens]`

vaslabs:
Thanks for the response.

I've just read #1019 , this is a bit different than zip (I think) as it gives all the syntax from cats (including mapN, map2 and anything derived from semigroupal and invariant)

But what you mention is a bit worrying:

> One thing to note is that tupled can break the optics which is fine, it only happens if the Lenses intersect.

Please excuse my ignorance here , but what does it mean for Lenses to intersect? Could you give some pointers to see if I manage to replicate this breaking with a test?

julien-truffaut:
Sorry I didn't give you enough context, I recently made a presentation where I show the potential issue with `zip` , slide 72 to 79: https://blog.fp-tower.com/monocle-3-a-peek-into-the-future/#72

IMO, `tupled` is completely fine, but I don't think `mapN` is possible, here is a counter example:
```scala
case class Point(x: Int, y: Int)

val sum: Lens[Point, Int] = (Focus[Point](_.x), Focus[Point](_.y)).map2(_ + _)

sum.get(Point(3, 5)) // that's fine
// res = 8

[...]

PR #1109, opened 2021-03-01
Feature/cats instances

vaslabs:
Following #1087 

Provides Invariant and Semigroupal cats instances for Lens.

## Motivation
Get some out of the box syntactic sugar from cats

I added a few Laws tests for Lens and for Getter (noticed it had Arrow already).

### Intersecting lenses

I've added a test to demonstrate a potential UX(?) issue discussed with @julien-truffaut 

```scala
test("tupled breaks get what you replace if you tuple the same/conflicting lenses") {
    val zippedLenses: Lens[Point, (Int, Int)] = (x, x).tupled
    assertEquals(
      zippedLenses.replace((2, 3))(Point(5, 6)),
      Point(3, 6)
    ) //2 disappeared

  }
  ```

Should I drop it in a doc somewhere?

### Laws tests
The tests rely on a generator method provider from cats to sample test the Eq instance (because we cannot prove two Lens instances are equal until we generate equal objects that they operate on. So this generates some tuple `S -> A` , creates a Lens on `S` and sets `A`, and yields (via via `lens.replace` and `lens.get` tuple) a new `S -> A` to be derived Equal from an `Eq[(S, A)]` instance.

Bit of coding gymnastics so I'd be a bit sceptical maintaining this, especially since sampleEq is deprecated.

Please let me know your thoughts.

### CODE REVIEW

File: test/shared/src/test/scala/monocle/LensInstancesSpec.scala [hunk 1]
```diff
@@ -0,0 +1,79 @@
...
+  implicit val arbSampleLensBool: Arbitrary[Lens[Sample, Boolean]] = Arbitrary(
+    Gen.const(Lens[Sample, Boolean](_.c)(b => s => s.copy(c = b)))
+  )
+
+  checkAll("Lens.InvariantLaws", InvariantTests[Lens[Sample, *]].invariant[Char, Int, Boolean])
+
+  checkAll("Lens.SemigroupalLaws", SemigroupalTests[Lens[Sample, *]].semigroupal[Char, Int, Boolean])
+
+  //Explicit tests for clarity
+  test("conflicting lenses are left associative") {
```
julien-truffaut, line 37: what do you mean by left associative?

julien-truffaut:
Thanks @vaslabs 
The cats spec seems to be a pain to setup, not sure if it is worth it. Maybe we could only zip two lenses (with different fields) and run the `LensTests`. 

It is good idea to have a unit test to show where it could break the Lens property.

test/shared/src/test/scala/monocle/LensInstancesSpec.scala
vaslabs: maybe wrong language, I meant the right setter operation overwrites the left one, so if they conflict the right most operation (lens replace) result is the result we get

vaslabs:
> Thanks @vaslabs
> The cats spec seems to be a pain to setup, not sure if it is worth it. Maybe we could only zip two lenses (with different fields) and run the `LensTests`.
> 
> It is good idea to have a unit test to show where it could break the Lens property.

I agree, only had it in initial PR as proof

test/shared/src/test/scala/monocle/LensInstancesSpec.scala
vaslabs: I might need to find something better here, the API signature does not point to what's going on
```
zippedLenses.replace(('b', 'c'))
```
Any suggestions?

vaslabs: will probably make a better test with fold or foldMap

vaslabs: I've cleaned it up, let me know what you think

julien-truffaut (approved):
LGTM, thanks for your work!

PR #1110, opened 2021-03-03
add as extension method to optics in Scala 2 and 3

julien-truffaut:
For Scala 2, I added extension methods in the macro module which basically do `optic.andThen(GenPrism[To, CastTo])`.

For Scala 3, I added the same extension methods but in the core module. I could have implemented it by forwarding it to `Focus(_.as[CastTo])` but I am worried about the documentation. 

I mean for `Focus` it makes sense to use `transparent` because the return type may vary but here we know it before hand. Maybe one of you know a way to do it? Admittedly, I am just copy/pasting and tweaking code but I have no good understanding of the macro APIs.

TODO in another PR, do the same for `ApplyOptics`.

### DIFF
--- a/core/shared/src/main/scala-3.x/monocle/syntax/MacroSyntax.scala
+++ b/core/shared/src/main/scala-3.x/monocle/syntax/MacroSyntax.scala
@@ -5,36 +5,32 @@ import monocle.{Focus, Fold, Iso, Optional, Prism, Setter, Traversal}
 import scala.quoted.{Expr, Quotes, Type}
 
 trait MacroSyntax {
-  implicit def toMacroPrismOps[From, To](optic: Prism[From, To]): MacroPrismOps[From, To] = new MacroPrismOps(optic)
-  implicit def toMacroOptionalOps[From, To](optic: Optional[From, To]): MacroOptionalOps[From, To] = new MacroOptionalOps(optic)
-  implicit def toMacroTraversalOps[From, To](optic: Traversal[From, To]): MacroTraversalOps[From, To] = new MacroTraversalOps(optic)
-  implicit def toMacroSetterOps[From, To](optic: Setter[From, To]): MacroSetterOps[From, To] = new MacroSetterOps(optic)
-  implicit def toMacroFoldOps[From, To](optic: Fold[From, To]): MacroFoldOps[From, To] = new MacroFoldOps(optic)
-}
 
-class MacroPrismOps[From, To](private val optic: Prism[From, To]) extends AnyVal {
-  inline def as[CastTo <: To]: Prism[From, CastTo] =
-    optic.andThen(GenPrism[To, CastTo])
-}
+  extension [From, To] (optic: Prism[From, To]) {
+    inline def as[CastTo <: To]: Prism[From, CastTo] =
+      optic.andThen(GenPrism[To, CastTo])
+  }
 
-class MacroOptionalOps[From, To](private val optic: Optional[From, To]) extends AnyVal {
-  inline def as[CastTo <: To]: Optional[From, CastTo] =
-    optic.andThen(GenPrism[To, CastTo])
-}
+  extension [From, To] (optic: Optional[From, To]) {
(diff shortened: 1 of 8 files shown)


core/shared/src/main/scala-3.x/monocle/syntax/MacroSyntax.scala
```diff
@@ -0,0 +1,51 @@
+package monocle.syntax
+
+import monocle.{Focus, Fold, Iso, Optional, Prism, Setter, Traversal}
+
+import scala.quoted.{Expr, Quotes, Type}
+
+trait MacroSyntax {
```
kenbot, line 7: Since we are in the `scala-3.x` bucket, can't we use the proper `extension` feature? Implicits are deprecated in Scala 3.

core/shared/src/main/scala-3.x/monocle/syntax/MacroSyntax.scala
```diff
@@ -0,0 +1,51 @@
...
+  implicit def toMacroPrismOps[From, To](optic: Prism[From, To]): MacroPrismOps[From, To] = new MacroPrismOps(optic)
+  implicit def toMacroOptionalOps[From, To](optic: Optional[From, To]): MacroOptionalOps[From, To] = new MacroOptionalOps(optic)
+  implicit def toMacroTraversalOps[From, To](optic: Traversal[From, To]): MacroTraversalOps[From, To] = new MacroTraversalOps(optic)
+  implicit def toMacroSetterOps[From, To](optic: Setter[From, To]): MacroSetterOps[From, To] = new MacroSetterOps(optic)
+  implicit def toMacroFoldOps[From, To](optic: Fold[From, To]): MacroFoldOps[From, To] = new MacroFoldOps(optic)
+}
+
+class MacroPrismOps[From, To](private val optic: Prism[From, To]) extends AnyVal {
```
kenbot, line 15: Why do we need an extension per optic type, when have a class hierarchy now?

kenbot (commented):
Changes should be made to Focus too:
- The type signature of the `as` keyword should be changed to be consistent with `[From, To <: From]`, which is possible now in RC1
- The implementation of the `AsGenerator` in `Focus` should delegate to this external operator functionality, instead of using its own implementation.

core/shared/src/main/scala-3.x/monocle/syntax/MacroSyntax.scala
julien-truffaut: That's for narrowing the output type `lens.as` is an `Optional` but `getter.as` is a `Fold`. There might be another way to do it with transparent,  I am not sure.

core/shared/src/main/scala-3.x/monocle/syntax/MacroSyntax.scala
julien-truffaut: Good idea!

core/shared/src/main/scala-3.x/monocle/syntax/MacroSyntax.scala
julien-truffaut: Btw, you will note that there is not an extension per Optic type, for example, there is not one `Getter`, `Iso`, or `Lens` as we don't need them to get a more precise return type.

julien-truffaut:
@kenbot I tried to change the definition of `as` extension method but it broke the parser. As far as I understand, it is because the `CastTo <: A` constraint add an implicit `ev: CastTo <:< A`.
```
[error]    |Unexpected code structure: TypeApply(Apply(TypeApply(Select(Ident(evidence$2),as),List(TypeTree[TypeRef(ThisType(TypeRef(ThisType(TypeRef(NoPrefix,module class focus)),class FocusAsTest)),trait Food)])),List(Select(Ident(_$2),mainIngredient))),List(Ident(Apple)))
```

kenbot (approved):
Merge away, I'll send the Focus changes in a new PR

Issue #1032, opened 2021-02-08
[Focus] Generate Isos rather than Lenses for single-argument case classes


julien-truffaut:

kenbot:
This isn't going to fly, Isos are less capable than lenses. We can't just have, say `case class UserId(id: Int)` and then instead of `replace` you can only do `reverseGet` by surprise. I don't think we should automatically generate them at all from Focus.

julien-truffaut:
@kenbot `Iso` have a `replace` method https://github.com/optics-dev/Monocle/blob/master/core/shared/src/main/scala/monocle/Iso.scala#L79

`Iso extends Lens`, they are strictly more powerful.

kenbot:
Oh, I've completely confused myself! You're right of course. All good

julien-truffaut:
> Oh, I've completely confused myself! You're right of course. All good

I also get confused regularly!

PR #1111, opened 2021-03-05
Focus on case classes with one fields generates an Iso

kenbot:
Selecting a lone case class field in `Focus` now generates an Iso, resolving #1032.

This was way harder than I thought it would be. 

Of note:
- I have represented it as its own feature bucket `selectonlyfield`, with its own `FocusAction.SelectOnlyField`.
- Generating the Iso required calling `apply` on the companion object, which needed some TASTy magic that I couldn't figure out, I had to get help from the Dotty gitter.
- We can now perceive that there are two main types of features, keyword features and field selection features. Most of the code that was in the `FieldSelectParser` is common between field selection features, and has been pulled out into `SelectParserBase`. The existing keyword matching machinery in `ParserBase` has been pulled out into `KeywordParserBase`.
- For the time being, I am just keeping one level of features, without nesting them into `keyword` and `select`. I generally prefer branching for intent rather than mechanism.
- I changed the generators to just take their `FocusAction` object, rather than all of their fields. This is because the action contents were being uselessly splattered out and recollected to pass to the generators in `GeneratorLoop`. This was noisy, brittle and a drag on refactoring.


core/shared/src/main/scala-3.x/monocle/internal/focus/features/at/AtGenerator.scala
```diff
@@ -8,9 +8,12 @@ private[focus] trait AtGenerator {
 
   import macroContext.reflect._
 
-  def generateAt(fromType: TypeRepr, toType: TypeRepr, index: Term, atInstance: Term): Term = 
+  def generateAt(action: FocusAction.KeywordAt): Term = {
```
julien-truffaut, line 11: much simpler this way

julien-truffaut (approved):
Awesome work @kenbot

PR #1112, opened 2021-03-06
Brings `as` into line with the other operators wrt Focus

kenbot:
- Thanks to Scala 3 RC1, adds a `From` parameter to the Focus `as` keyword, constrained by a type bounds. This is now the same signature as the external `as` operator:
```
extension [From] (from: From)
      def as[CastTo <: From]: CastTo
```
-  Adds a case to the `FocusKeyword` pattern matcher to correctly interpret an extension method with both inferred and direct type parameter lists.
- Renamed `GenPrism` to `AsPrism`. It is no longer analogous to the deprecated `GenXXX` features, and should not forward to `Focus`. It is an implementation detail for the new and supported `as` operator.
- Changed the `AsGenerator` implementation to directly use `AsPrism`, rather than have its own operation. This is now the same as the other features, which all generate code that delegates to the external operator. We are using the implementation class `AsPrism` rather than the direct `as` operator because the external one is only accessible as an optic extension method. The way Focus works, each optic is generated on its own and composed together later, so we can't call an extension method on the last thing generated.
- Removed the `CastTo <: To` constraint in the `AsPrism` implementation, because proving the relationship was hard inside the Focus macro. It is sufficient to enforce the constraint in the public interface.
- Added some tests to confirm that the Focus and external `as` operators commute with respect to each other, just like the other operators

### DIFF
--- a/core/shared/src/main/scala-3.x/monocle/internal/focus/features/as/AsGenerator.scala
+++ b/core/shared/src/main/scala-3.x/monocle/internal/focus/features/as/AsGenerator.scala
@@ -2,6 +2,7 @@ package monocle.internal.focus.features.as
 
 import monocle.Prism
 import monocle.internal.focus.FocusBase
+import monocle.syntax.AsPrism
 
 private[focus] trait AsGenerator {
   this: FocusBase => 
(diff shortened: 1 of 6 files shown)


core/shared/src/main/scala-3.x/monocle/internal/focus/features/as/AsGenerator.scala
```diff
@@ -12,9 +13,7 @@ private[focus] trait AsGenerator {
     import action.{fromType, toType}
 
     (fromType.asType, toType.asType) match {
-      case ('[f], '[t]) => '{ 
-        Prism[f, t]((from: f) => if (from.isInstanceOf[t]) Some(from.asInstanceOf[t]) else None)
-                   ((to: t) => to.asInstanceOf[f]) }.asTerm
+      case ('[f], '[t]) => '{ AsPrism[f, t] }.asTerm
```
julien-truffaut, line 16: Nice

julien-truffaut (approved):
Great!

PR #1113, opened 2021-03-06
create type hirearchy and aliases that matches standard optics

julien-truffaut:
Since we are breaking compatibility, it is also the moment to rename `ApplyOptics` if we find a better name.

Maybe `PartiallyAppliedXXX`, e.g. `PartiallyAppliedLens` but it is very long :(


julien-truffaut:
@kenbot I updated all named to `Applied`

kenbot (approved):
lgtm

PR #1114, opened 2021-03-06
Deprecate first, second, left, right, split. Move choice to cats instance

julien-truffaut:
I doubt these methods see much use. We implemented them because we could but it is pain to mainten and it pollutes the interface.


cquiroz (approved):
Looks fine, I don't recall ever using them

kenbot (approved):
Lgtm

PR #1115, opened 2021-03-06
Deprecate all except Lens codiagonal

julien-truffaut:
We don't need the other ones now that we have inheritance between optics.


cquiroz (approved):
Looks great, a lot simpler

PR #1120, opened 2021-03-11
offer same import for Scala 2 and 3

julien-truffaut:
This PR moves `monocle.syntax.all` and `monocle.Monocle` (I forgot we had it) from `core` to `macros` module for Scala 2 only.

This way we can use the same import for Scala 2 and Scala 3 even though it is defined in different module, which should simplify the documentation and user migration.
```scala
import monocle.syntax.all._
```


kenbot (approved):
lgtm

PR #1123, opened 2021-03-12
`as` for AppliedOptics

julien-truffaut:
1. Fix a bug in `AppliedIso` constructor thanks to @kenbot 
2. defined an alias for `AppliedGetter` and `AppliedFold` in the top level `monocle` package


kenbot (approved):
lgtm

PR #1125, opened 2021-03-14
Field syntax for optics

julien-truffaut:
Add a `field(lambda)` on all `Optics` such that:

```scala
case class User(name: String, address: Address)
case class Address(streeetNumber: Int, postcode: String)

Focus[User]()
  .field(_.address)
  .field(_.streetNumber) == Focus[User](_.address.streetNumber)
```

It is a bit redundant but this method is meant for updating existing optics:

```scala
val user: Lens[State, User] = ... // suppose we have it defined

user.field(_.name) == user.andThen(Focus[User](_.name))
```

### DIFF
--- a/macro/src/main/scala-2.x/monocle/macros/syntax/AppliedFocusSyntax.scala
+++ b/macro/src/main/scala-2.x/monocle/macros/syntax/AppliedFocusSyntax.scala
@@ -27,16 +27,6 @@ class GenAppliedLensOpsImpl(val c: blackbox.Context) {
         c.abort(c.enclosingPosition, s"Invalid prefix tree ${show(t)}")
     }
 
-    lambda match {
-      // _.field
-      case Expr(Function(List(ValDef(_, _, _, EmptyTree)), Select(Ident(_), _))) => ()
-      case _ =>
-        c.abort(
-          c.enclosingPosition,
-          s"Illegal field reference ${show(lambda.tree)}; please use _.field... instead"
-        )
-    }
-
     c.Expr[AppliedLens[A, C]](q"""
       _root_.monocle.syntax.AppliedPLens(
         $subj,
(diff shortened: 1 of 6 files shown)


macro/src/main/scala-2.x/monocle/macros/syntax/AppliedFocusSyntax.scala
```diff
@@ -27,10 +27,20 @@ class GenAppliedLensOpsImpl(val c: blackbox.Context) {
         c.abort(c.enclosingPosition, s"Invalid prefix tree ${show(t)}")
     }
 
+    lambda match {
+      // _.field
+      case Expr(Function(List(ValDef(_, _, _, EmptyTree)), Select(Ident(_), _))) => ()
```
julien-truffaut, line 32: This is really ugly to duplicate code from `GenLens` macro but I couldn't find out how to pass a Boolean flag ...

I don't think it matters too much since it is for Scala 2 and we don't plan too touched macro further in this version of Scala.

macro/src/main/scala-2.x/monocle/macros/syntax/MacroSyntax.scala
```diff
@@ -76,4 +108,27 @@ class MacroAsOpsImpl(val c: blackbox.Context) {
...
+
+  def field_impl[Optic[_, _], From, To: c.WeakTypeTag, Next](lambda: c.Expr[To => Next]): c.Expr[Optic[From, Next]] = {
+    import c.universe._
+
+    val subj = c.prefix.tree match {
+      case Apply(TypeApply(_, _), List(x)) => x
+      case t                               => c.abort(c.enclosingPosition, s"Invalid prefix tree ${show(t)}")
+    }
+
+    lambda match {
```
julien-truffaut, line 120: This duplicate code from GenLens macro, I couldn't find out how to pass a Boolean flag ...

I don't think it matters too much since it is for Scala 2 and we don't plan to touch macro further in this version of Scala.

julien-truffaut:
need more discussion, closing for now

PR #1157, opened 2021-05-18
remove sbt-dotty plugin

xuwei-k:
https://github.com/lampepfl/dotty/blob/08dda0de2138b5f7be6c0901e82634e16804f468/sbt-dotty/src/dotty/tools/sbtplugin/DottyPlugin.scala#L188-L189


cquiroz (approved):
LGTM

PR #1184, opened 2021-07-26
Refocus feature

### LINKED ISSUE #1056: [Focus] Mix Focus and standard optics
`Focus` goal is to provide the best syntax for the most common use cases, but we know we will never cover all scenarios.

 So at some point, users may need to mix custom optics with `Focus`, currently they can use `andThen`:

```scala
Focus[Foo](_.some.bar).andThen(customOptic)

customOptic.andThen(Focus[Foo](_.some.bar))
```

While, it is not too bad, it might be worth adding two methods:
1. `andThen` inside of `Focus`, e.g.  `Focus[Foo](_.some.bar.andThen(customOptic))`
2. `focus` on optics/applyOptics, e.g. `customOptic.foucs(_.some.bar)`

I am not sure these methods are worth the maintaince cost. We can also wait for `3.0` to be released and see the common use cases people have. We don't need to solve everything at once.

cc @yilinwei @kenbot

kenbot:
- Adds extension method `refocus` to all optics & applied optics, which is equivalent to `optic.andThen(Focus[Foo](_.....))`.
- Resolves #1056
- Must be an extension method rather than a direct method on optics, so as to cleanly separate the macro code.
- Astonishingly, can treat all optics and applied optics uniformly with only one simple implementation.
- Privately defines a `type AnyOptic[S,A] = Setter[S,A] | Fold[S,A] | AppliedSetter[S,A] | AppliedFold[S,A]`. Should this be a public definition? Probably not, because its usage is not likely to be meaningful beyond this macro, and would be highly susceptible to change.
- Big fat test suite testing many combinations of optic types, and especially `Getter` and `Fold`, since they are special cases, because their `andThen` method only has one type parameter.
- Have not added a Scala 2 equivalent, corresponding to the single-field Focus functionality supported in Scala 2. It makes sense to add this, but perhaps that can be a different ticket

### DIFF
--- a/core/shared/src/main/scala-3.x/monocle/Focus.scala
+++ b/core/shared/src/main/scala-3.x/monocle/Focus.scala
@@ -1,10 +1,10 @@
 package monocle
 
-import monocle.syntax.AppliedFocusSyntax
+import monocle.syntax.{AppliedFocusSyntax, ComposedFocusSyntax}
 import monocle.internal.focus.FocusImpl
 import monocle.function.{Each, At, Index}
 
-object Focus extends AppliedFocusSyntax {
+object Focus extends AppliedFocusSyntax with ComposedFocusSyntax {
 
   sealed trait KeywordContext {
     extension [From] (from: From)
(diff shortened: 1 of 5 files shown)


julien-truffaut:
Awesome work @kenbot 
For Scala 2, maybe we can resurect https://github.com/optics-dev/Monocle/pull/1125

I don't have a strong opinion about `AnyOptic` (public or private)

julien-truffaut (approved):

PR #1185, opened 2021-08-02
Add GenIso to Scala 3

### LINKED ISSUE #1043: GenIso.fields for Scala 3
Generate an `Iso` between a case class and a tuple of all its fields.

https://github.com/optics-dev/Monocle/blob/12c89e1c5bcc95e121f1f548a28e463e25f12a15/macro/src/main/scala-2.x/monocle/macros/GenIso.scala#L23

For example

```scala
case class User(name: String, age: Int)

GenIso.fields[User]
// Iso[User, (String, Int)]
```

japgolly:
Closes #1043


julien-truffaut:
Sorry I missed this PR. LGTM!

macro/src/main/scala-3.x/monocle/macros/GenIso.scala
```diff
@@ -0,0 +1,98 @@
...
+
+import monocle.Iso
+import scala.language.`3.0`
+import scala.deriving.*
+import scala.quoted.*
+
+object GenIso {
+
+  /** Generate an [[Iso]] between a case class `S` and its unique field of type `A`. */
+  inline def apply[S <: Product, A](using m: Mirror.ProductOf[S]): Iso[S, A] =
```
kenbot, line 11: `Focus[Foo](_.onlyField)` already generates an Iso, I don't think we need a second way to do this

macro/src/main/scala-3.x/monocle/macros/GenIso.scala
```diff
@@ -0,0 +1,98 @@
...
+          Iso[S, A](f)(g)
+        }
+
+      case '{ $m: Mirror.ProductOf[S] {type MirroredElemTypes = a} } =>
+        report.throwError(s"Can't generate an Iso[${Type.show[S]}, ${Type.show[A]}] because it's fields are ${TypeRepr.of[a].show} ")
+    }
+  }
+
+  /** Generate an [[Iso]] between an object `S` and `Unit`. */
+  inline def unit[S <: Product](using m: Mirror.ProductOf[S]): Iso[S, Unit] =
```
kenbot, line 31: Ah interesting! I think this is useful, but it's not very obvious reading this that it's only an Iso with empty products, like case objects. Would it be better to use (iirc) the new Singleton trait?

kenbot (commented):
Great work here @japgolly! I have only some nitpicks.

Aside to @julien-truffaut : we normally have constructors in companion objects, shouldn't we do that here too? 

We used to have `GenXXX` for constructors with metaprogramming, but I don't see any reason to reintroduce this convention, users shouldn't care about the impl and they don't need one more name to remember.

macro/src/main/scala-3.x/monocle/macros/GenIso.scala
julien-truffaut: Maybe we should add a `deprecated` annotation to point to focus.

PR #1186, opened 2021-08-03
Avoid indirect call to AsPrismImpl

nicolasstucki:
* Avoid access to private module in quotes (see https://github.com/lampepfl/dotty/issues/13215)
* Make code more preformant


nicolasstucki:
Also related to https://github.com/lampepfl/dotty/pull/13188

julien-truffaut (approved):

julien-truffaut:
Thanks @nicolasstucki

Issue #1183, opened 2021-07-22
Compilation error when importing cats.syntax.all


cquiroz:
It is common that we use monocle and cats together, I noted some of the examples fail to compile when I also have imported `cats.syntax.all._`

It is hard to tell if this is an actual error in Monocle or we should recommend some particular way to do the cats import

See
https://scastie.scala-lang.org/cquiroz/zYAqjfE5R4Gee20gpBcy9Q/5

kenbot:
Here's the culprit; Cats' Representable has an `index` extension method. https://github.com/typelevel/cats/blob/e0784fc0d95b9c4eed0bb1ddce6c69e3fd9a4a4f/core/src/main/scala/cats/syntax/representable.scala

Git blame says it was added 3 years ago. We're part of the Typelevel/Cats family of stuff aren't we @julien-truffaut? Is there a line of conversation among the maintainers about having roughly compatible & non-colliding conventions? 

We can probably circumvent this by implementing `index` as a direct method on optics instead of a syntax extension.

kenbot:
While `index` has been in Monocle since forever, it's only in Monocle 3 that we added the extension method as a way of mitigating the type inference penalty for overloading `andThen`, so I guess we're the ones being bad neighbours! 😱  Cats is our designated FP library dependency, so it seems pretty important that our syntax extensions get along with it.

julien-truffaut:
Arf I thought we would catch these sorts of issues in our tests.

@kenbot Yes, we are part of Typelevel and most importantly, we depend on cats, so we should avoid these situations.

Let's check if it is possible to implement `index` as a method of all optics. Otherwise, we could rename it to `idx` or something similar. WDYT?

kenbot:
I think a breaking change is bad for users, and `idx` is a worse name and not in keeping with the direction of avoiding cryptic symbols and abbreviations. So I think making `index` a direct method is a good solution. 

Which begs the question... what about the other extension methods - `at`, `some`, etc. Should they be regular methods too? What is the rationale for some being extensions and others direct methods?

julien-truffaut:
I believe I used extension methods because the signature on the optics class was extremely complex (lots of `A =:= B`, `S =:= T` etc). There were some corss compilation issues with Scala 3. 
It is worth giving it a another try though.

kenbot:
We could keep them in a mixin trait and get the optics to extend it?

kenbot:
[...]

PR #1192, opened 2021-08-21
index is now a direct method, for better Cats compat

kenbot:
Fix for #1183 

- Index is now a direct method on all optics, instead of an extension method
- new private method on all polymorphic optics `adaptMono`

### DIFF
--- a/core/shared/src/test/scala/monocle/CatsCompatTest.scala
+++ b/core/shared/src/test/scala/monocle/CatsCompatTest.scala
@@ -1,53 +1,52 @@
-
 package monocle
 
 final class CatsCompatTest extends munit.FunSuite {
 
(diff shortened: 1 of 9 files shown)


core/shared/src/test/scala/monocle/CatsCompatTest.scala
```diff
@@ -0,0 +1,53 @@
+
+package monocle
+
+final class CatsCompatTest extends munit.FunSuite {
+
+  test("Setter index doesn't conflict with Cats") {
+    import cats.syntax.all._
+    val setter = Setter.id[List[Int]]
```
cquiroz, line 8: Cool

cquiroz (approved):
Thanks that looks great

PR #1197, opened 2021-08-28
Some housekeeping for new GenIso.fields functionality

kenbot:
- Deprecated the public Scala 3 `GenIso` methods
- Moved the new Scala 3-compatible `GenIso.fieldsTuple` functionality into the `core` module
- Moved the implementation of the `AsPrism` macro into its own file in `monocle.internal` to set a better pattern and get implementations out of `MacroSyntax`
- Created a new file `IsoFields` in this same pattern with the `fieldsTuple` macro implementation, and added a forwarding method in `MacroSyntax` that adds a `fields` extension on to `object Iso`

Of note:
- I had to remove @japgolly 's delicious discipline/law tests, because the `core` module cannot depend on other modules, everything else depends on `core`. I replaced it with a meat & potatoes example test, which in this case I think will be fine. 
- I couldn't define `Iso.fields` directly on `Iso`, because it uses a Scala 3 macro, and the core files like `Iso` have to still compile against Scala 2. The extension method `fields` requires a syntax import however, so this detail unfortunately leaks into the user's awareness & workflow.


macro/src/main/scala-3.x/monocle/macros/GenIso.scala
```diff
@@ -28,6 +29,7 @@ object GenIso {
   }
 
   /** Generate an [[Iso]] between an object `S` and `Unit`. */
+  @deprecated("no replacement", since = "3.1.0")
```
julien-truffaut, line 32: You don't think this method is useful? I don't have a strong optinon.

julien-truffaut:
Looks great!

macro/src/main/scala-3.x/monocle/macros/GenIso.scala
kenbot: Yeah I don't really have a strong opinion either. I can retain it by putting another extension method on `Iso` in core and making this forward to it, maybe I'll do that.

kenbot: `Iso.fields` automatically gives us an `Iso[Foo, EmptyTuple]` for case objects. We can implement the `Unit` on top of this rather than creating another macro

julien-truffaut: ah that's perfect, we can simply deprecate in favor of `Iso.fields`. If someone needs it, we can add an `Iso[ EmptyTyple, Unit]` somewhere.

julien-truffaut:
No all great. Sorry I forgot to merge.

Issue #1259, opened 2022-02-05
Compilation error with context bounds


NTPape:
Hello there!

Compilation of the following code will fail

```
import monocle.Lens
import monocle.macros.GenLens

trait Foo[T]
trait Bar[T]

case class A[T: Foo](s: A.S[T]) {
  private val lens: Lens[A.S[T], Bar[T]] = GenLens[A.S[T]](_.bar)
}

object A {
  case class S[T: Foo](bar: Bar[T])
}
```

in Scala 3 with the following error
```
[error] 8 |  private val lens: Lens[A.S[T], Bar[T]] = GenLens[A.S[T]](_.bar)
[error]   |                                           ^^^^^^^^^^^^^^^^^^^^^^
[error]   |Exception occurred while executing macro expansion.
[error]   |java.lang.Exception: Expected an expression. This is a partially applied Term. Try eta-expanding the term first.
```

and in Scala 2.13 with
```
[error] could not find implicit value for evidence parameter of type Foo[T]
[error]   private val lens: Lens[A.S[T], Bar[T]] = GenLens[A.S[T]](_.bar)
[error]                                                           ^
```

Interestingly, the last version that appears to work is 1.5.1-cats on Scala 2.11 (but not on Scala 2.12.)

NTPape:
I had a look into the issue in **Scala 2.13** and as far as I can see the issue is located in the lens's modifyF method because it has a context bound for cats.Functor. Due to macro magic, the Functor evidence is named like it were the first evidence in the source file, i.e. like the evidence for Foo of the enclosing class A, so the evidence for Foo is shadowed and hence unavailable in modifyF's function body where it is needed for copy.

That means one workaround for the issue in **Scala 2.13** is putting another class with a context bound above the case class to rename the evidences below it, like this:

```
import monocle.Lens
import monocle.macros.GenLens

trait Foo[T]
trait Bar[T]

private class IncrementAllEvidencesBelow[T: Foo]()

case class A[T: Foo](s: A.S[T]) {
  private val lens: Lens[A.S[T], Bar[T]] = GenLens[A.S[T]](_.bar)
}

object A {
  case class S[T: Foo](bar: Bar[T])
}
```

julien-truffaut:
Thanks for investigating. Do you think it will solve the problem if we change https://github.com/optics-dev/Monocle/blob/master/macro/src/main/scala-2.x/monocle/macros/internal/Macro.scala#L100

from:
```
[...]

PR #1265, opened 2022-02-20
Fix GenLens issue with evidence shadowing

NTPape:
Will (mostly) solve #1259 for Scala 2.13.


julien-truffaut:
Great. Thanks @NTPape

PR #1267, opened 2022-02-25
Fix compilation of lenses with context bounds in Scala 3

### LINKED ISSUE #1259: Compilation error with context bounds

NTPape:
Fixes #1259


NTPape:
I did exactly what the error message said: try to eta expand the Term if it's not an Expr yet. This should have no influence on working code because for all code where it already compiles, i.e. asExpr succeeds, isExpr is already true and hence eta expanding would be skipped.

NTPape:
I had to move the `S` case class out of the companion in the test due to a bug in the dotty compiler having trouble finding the (correct) `companionModule` of locally defined classes. I think it's covered by this issue: https://github.com/lampepfl/dotty/issues/14508

julien-truffaut:
Thanks @NTPape ! @kenbot would you mind to review this PR when you have time?

NTPape:
Thanks for the detailed explanation, @kenbot!

I gave it a quick go and the issue I'm running into is lifting the FocusResult of a splice into the surrounding expression. I.e. `etaExpandIfNecessary` now returns a `FocusResult[Term]` and now I'm trying to resolve the setter in the parser like so:

```Scala
(fromType.asType, toType.asType) match {
  case ('[f], '[t]) =>
    '{
      (to: t) => ${
        etaExpandIfNecessary(
          Select.overloaded(companion, "apply", fromTypeArgs, List('{ to }.asTerm))
        ).asExprOf[f]
      }
    }.asTerm)
}
```

Is there a nice way to lift the FocusResult into the surrounding expression, or is there another way how to get a parameter for the overload resolution?
(I haven't tried it yet but I guess one hacky solution would be to throw an exception, catch it, and convert it back into a FocusResult.)

NTPape:
Sorry, my last comment was a bit too concise and without having had uploaded the code yet, it was unnecessarily difficult to understand my question and to follow the current state of my attempt. I have now uploaded a working version.

As far as I understand, it is necessary to move the setter creation completely into the parser then, because only in the expression `(to: To) => fromCompanion.apply(to)` the type parameters of `from` become initialized in a way which allows for implicit search for values which needs these type parameters. (One cannot search for Foo[T] without knowing what T is, but in the expression above it becomes A.this.T in the spec for example, which can be found.)

I also used the hacky FocusResult lifting via throwing exceptions for now because I'm not aware of a nicer way.

Is this going in the right direction or am I on the wrong track here? Thanks for your help!

NTPape:
I did the split now and merged the Select(Only)Field parsers and generators because they are so similar.

Can you have a look again?

PR #1285, opened 2022-06-15
Implemented generic `Index` for various Scala 2 sequences

skozlov:
Makes implicit `Index` available not only for `List`, `LazyList`, and `Vector`, but for many other sequences.

I also believe that this approach can be used in different classes (e.g. `At` for `Map`s).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -100,8 +100,8 @@ lazy val catsFree          = Def.setting("org.typelevel" %%% "cats-free" % catsV
 lazy val catsLaws          = Def.setting("org.typelevel" %%% "cats-laws" % catsVersion)
 lazy val alleycats         = Def.setting("org.typelevel" %%% "alleycats-core" % catsVersion)
 lazy val shapeless         = Def.setting("com.chuusai" %%% "shapeless" % "2.3.9")
-lazy val refinedDep        = Def.setting("eu.timepit" %%% "refined" % "0.9.29")
-lazy val refinedScalacheck = Def.setting("eu.timepit" %%% "refined-scalacheck" % "0.9.29" % "test")
+lazy val refinedDep        = Def.setting("eu.timepit" %%% "refined" % "0.10.0")
+lazy val refinedScalacheck = Def.setting("eu.timepit" %%% "refined-scalacheck" % "0.10.0" % "test")
 
 lazy val discipline      = Def.setting("org.typelevel" %%% "discipline-core" % "1.5.1")
 lazy val munit           = Def.setting("org.scalameta" %% "munit" % "0.7.21" % Test)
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/monocle/function/Index.scala
```diff
@@ -45,6 +46,13 @@ object Index extends IndexFunctions {
   /** Std instances */
   /** *********************************************************************************************
     */
+  implicit def seqIndex[A, S[B] <: SeqOps[B, S, S[B]]]: Index[S[A], Int, A] =
+    Index(i =>
+      if (i < 0) Optional.void
+      else
+        Optional[S[A], A](s => Some(i).filter(s.indices.contains).map(s.apply))(a => s => Try(s.updated(i, a)).getOrElse(s))
```
julien-truffaut, line 53: is it faster to do `Some(i).filter(s.indices.contains)` than to catch the exception if the indice doesn't exist?

core/shared/src/main/scala/monocle/function/Index.scala
```diff
@@ -45,6 +46,13 @@ object Index extends IndexFunctions {
   /** Std instances */
   /** *********************************************************************************************
     */
+  implicit def seqIndex[A, S[B] <: SeqOps[B, S, S[B]]]: Index[S[A], Int, A] =
```
julien-truffaut, line 49: Do we need the implicit for List, Vector, etc? 
Can you also add a test for Seq.

julien-truffaut:
@skozlov Thanks, I used to be against using `Seq` but now, I think it is great if we can make `Index` for all `Seq`.
I added a few comments.

core/shared/src/main/scala/monocle/function/Index.scala
skozlov: Good question. It depends...
After thinking about it, I realized that `Try(s(i)).toOption` is the best default implementation for sequences as its only drawback in terms of performance is the exception handling overhead.
On the other hand, implementations of `IndexedSeqOps` are supposed to have efficient `indices` method, so I believe that it makes sense to create a corresponding `Index` in the future.

skozlov:
@julien-truffaut You're welcome. Thank you for the project and for the really quick review!

core/shared/src/main/scala/monocle/function/Index.scala
skozlov: > Do we need the implicit for List, Vector, etc?
OK, we don't need specific implementations for them.
But should we remove them? It would break backward compatibility.

skozlov: > The question is does Scala select the most precise implicit?

Yes, it does.

julien-truffaut:
Sorry for the delayed review. Thanks for your work!

skozlov:
@julien-truffaut, thank you for merging this PR. I'm really happy to start contributing to the Scala ecosystem!

PR #1328, opened 2022-11-27
Add stringToURI

sideeffffect:


core/jvm/src/main/scala/monocle/std/PlatformSpecificString.scala
```diff
@@ -1,10 +1,14 @@
 package monocle.std
 
-import java.net.URL
+import java.net.{URI, URL}
 import monocle.Prism
 import scala.util.Try
 
 private[std] trait PlatformSpecificStringOptics {
+  val stringToURI: Prism[String, URI] =
+    Prism((s: String) => Try(new URI(s)).toOption)(_.toString)
```
armanbilge, line 9: `URI` is also on Scala.js (and Native I assume).
https://github.com/scala-js/scala-js/blob/0020f756c960b18caf330c8f7a80ccc3800cbe2a/javalib/src/main/scala/java/net/URI.scala

sideeffffect:
Yep! All good
https://github.com/optics-dev/Monocle/blob/master/core/shared/src/main/scala/monocle/std/String.scala#L34

This was a very reflexive PR :joy:

PR #1413, opened 2023-12-19
Improve error messages for the `Focus` macro.

yilinwei:
A few random changes which are (mostly) forced by compiler changes.

1. The inferred type for inline lambas is different, so we get this awfully confusing error message currently:

```
Cannot generate Lens for field 'foo', because '_$1' is not a case class
```

2. We can now report a position error within a macro so the language servers should be able to point to the right part of the code.

I think I'll change the hierarchy of the `FocusError` a little, because some of the errors are internal i.e. `5xx` and some are `4xx`; mixing them is quite confusing.

### DIFF
--- a/core/shared/src/main/scala-3/monocle/internal/focus/ErrorHandling.scala
+++ b/core/shared/src/main/scala-3/monocle/internal/focus/ErrorHandling.scala
@@ -3,19 +3,52 @@ package monocle.internal.focus
 private[focus] trait ErrorHandling {
   this: FocusBase =>
 
-  def errorMessage(error: FocusError): String = error match {
-    case FocusError.NotACaseClass(fromClass, fieldName) =>
-      s"Cannot generate Lens for field '$fieldName', because '$fromClass' is not a case class"
+  def errorReport(error: FocusError): (String, Option[Position]) = error match {
+    case FocusError.NotACaseClass(fromClass, fieldName, pos) =>
+      (
+        s"Cannot generate Lens for field '$fieldName', because '$fromClass' is not a case class",
+        Some(pos)
+      )
     case FocusError.NotAConcreteClass(fromClass) =>
-      s"Expecting a concrete case class in the 'From' position; cannot reify type $fromClass"
+      (
+        s"Expecting a concrete case class in the 'From' position; cannot reify type $fromClass",
+        None
+      )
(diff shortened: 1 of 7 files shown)


core/shared/src/main/scala-3/monocle/internal/focus/features/SelectParserBase.scala
```diff
@@ -26,16 +26,24 @@ private[focus] trait SelectParserBase extends ParserBase {
...
-    // for some reason their field names (_1, _2, etc) have a space at the end, ie `_1 `.
-    def getTrimmedFieldSymbol(fromTypeSymbol: Symbol): Symbol =
-      fromTypeSymbol.memberFields.find(_.name.trim == fieldName).getOrElse(Symbol.noSymbol)
+  private val tupleFieldPattern = "^_[0-9]+$".r
+
+  def getFieldType(fromType: TypeRepr, fieldName: String, pos: Position): FocusResult[TypeRepr] = {
+    def getFieldSymbol(fromTypeSymbol: Symbol): Symbol = {
+      // We need to do this to support tuples, because even though they conform as case classes in other respects,
+      // for some reason their field names (_1, _2, etc) have a space at the end, ie `_1 `.
+      val f: String => String =
```
yilinwei, line 35: This fixes a silent error if we have something like

```scala case class Foo(x: Int, `x `: Int)``` since identifiers with leading/trailing spaces are not equivalent.

File: core/shared/src/main/scala-3/monocle/internal/focus/ErrorHandling.scala [hunk 2]
```diff
@@ -3,19 +3,52 @@ package monocle.internal.focus
 private[focus] trait ErrorHandling {
   this: FocusBase =>
 
-  def errorMessage(error: FocusError): String = error match {
-    case FocusError.NotACaseClass(fromClass, fieldName) =>
-      s"Cannot generate Lens for field '$fieldName', because '$fromClass' is not a case class"
+  def errorReport(error: FocusError): (String, Option[Position]) = error match {
```
yilinwei, line 6: The macro has better support for errors now, I need to see whether we can use the `-explain` tag.

julien-truffaut (approved):

julien-truffaut:
Thanks @yilinwei, it looks good to me!

PR #1414, opened 2023-12-21
Introduces non-case class getters.

### LINKED ISSUE #1034: [Focus] Non-case class getters
There's no reason not to allow plain old getter methods on non-case classes, generating a `Getter`. This is something that a user might reasonably guess is supported, and it seems unreasonably to make them bust out the Optics Family documentation to do it. 

eg `Focus[String](_.length) // Getter[String,Int]`

Original issue: https://github.com/optics-dev/Monocly/issues/39

yilinwei:
- Fixes #1034
- Fixes generation when case classes have overloads

### DIFF
--- a/core/shared/src/main/scala-3/monocle/internal/focus/features/GeneratorLoop.scala
+++ b/core/shared/src/main/scala-3/monocle/internal/focus/features/GeneratorLoop.scala
@@ -64,7 +64,6 @@ private[focus] trait GeneratorLoop {
 
     lens1.tpe.classSymbol match {
       case Some(classSym) =>
-
         // This is to select the correct 'andThen' method. The algorithm
         // essentially loops through and finds the 'lowest' subtype which
         // is related to the term on the right.
--- a/core/shared/src/main/scala-3/monocle/internal/focus/features/GeneratorLoop.scala
+++ b/core/shared/src/main/scala-3/monocle/internal/focus/features/GeneratorLoop.scala
@@ -9,7 +9,7 @@ import monocle.internal.focus.features.each.EachGenerator
 import monocle.internal.focus.features.at.AtGenerator
 import monocle.internal.focus.features.index.IndexGenerator
 import monocle.internal.focus.features.withdefault.WithDefaultGenerator
-import monocle.{Iso, Lens, Optional, Prism, Traversal}
+import monocle.{function, Iso, Lens, Optional, PIso, PLens, POptional, PPrism, PTraversal, Prism, Traversal}
 import scala.quoted.Type
 
 private[focus] trait AllFeatureGenerators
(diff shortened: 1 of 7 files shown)


core/shared/src/main/scala-3/monocle/internal/focus/features/GeneratorLoop.scala
```diff
@@ -48,11 +48,38 @@ private[focus] trait GeneratorLoop {
...
+    }
+
   private def composeOptics(lens1: Term, lens2: Term): FocusResult[Term] =
-    lens2.tpe.widen match {
-      // Won't yet work for polymorphism where A != B, nor for non-polymorphic optics Getter or Fold.
-      case AppliedType(_, List(_, toType2)) =>
-        Right(Select.overloaded(lens1, "andThen", List(toType2, toType2), List(lens2)))
-      case _ => FocusError.ComposeMismatch(lens1.tpe.show, lens2.tpe.show).asResult
+    try
+      lens2.tpe.widen match {
```
yilinwei, line 75: We should be able to do something a _little_ cleverer than this. There's a couple of newer APIs in dotty which means it's a little easier.

core/shared/src/main/scala-3/monocle/internal/focus/features/selectfield/SelectFieldGenerator.scala
```diff
@@ -18,13 +18,29 @@ private[focus] trait SelectFieldGenerator {
...
+      case SelectType.PublicField =>
+        (fromType.asType, toType.asType) match {
+          case ('[f], '[t]) =>
+            '{ Getter.apply[f, t]((from: f) => ${ generateGetter('{ from }.asTerm).asExprOf[t] }) }.asTerm
+        }
+      case SelectType.VirtualField =>
+        (fromType.asType, toType.asType) match {
+          case ('[f], '[t]) =>
+            '{
+              Getter.apply[f, t]((from: f) => ${ generateGetter('{ from }.asTerm).appliedToArgs(List()).asExprOf[t] })
```
yilinwei, line 40: To be honest, we should stamp the symbol in the parser so we don't do the resolution a second time - this currently doesn't work when we have clashing names like most Java classes.

core/shared/src/main/scala-3/monocle/internal/focus/features/SelectParserBase.scala
```diff
@@ -29,25 +29,58 @@ private[focus] trait SelectParserBase extends ParserBase {
...
-        case FieldType(possiblyTypeArg) => Right(swapWithSuppliedType(fromType, possiblyTypeArg))
-        case _                          => FocusError.CouldntFindFieldType(fromType.show, fieldName, pos).asResult
+        case Some(sym) => Right(sym)
+        case _         => FocusError.CouldntFindFieldType(fromType.show, fieldName, pos).asResult
       }
     }
   }
 
+  def paramAccessorType(tpe: TypeRepr, symbol: Symbol, fieldName: String): Option[TypeRepr] = {
+    symbol.methodMember(fieldName).map(each => println(each.paramSymss))
```
yilinwei, line 57: Remember to remove

yilinwei:
Closing in favour of more complete PR later.

PR #1574, opened 2026-04-11
Use wildcard type instead of whitebox cast in IsoFields

mbovel:
Replace the `whitebox` cast (`asInstanceOf[Expr[Iso[S, Tuple]]]`) with a wildcard return type using `PIso[S, S, ? <: Tuple, ? <: Tuple]`.

Since `IsoFields.apply` is a `transparent inline` method, the compiler still infers the precise tuple type at call sites. This avoids the unsafe cast.

The motivation is to prepare for future stricter checks in the Scala 3 compiler (https://github.com/scala/scala3/pull/25756). The current implementation exploits a missing check to generate an unsound cast.

Note: we use `PIso` (a trait) directly instead of the `Iso` type alias (defined as `type Iso[S, A] = PIso[S, S, A, A]`) because Scala 3 cannot reduce higher-kinded type aliases applied to wildcard arguments. Is that acceptable to use `PIso` instead of `Iso`?

### DIFF
--- a/core/shared/src/main/scala-3/monocle/internal/IsoFields.scala
+++ b/core/shared/src/main/scala-3/monocle/internal/IsoFields.scala
@@ -1,28 +1,23 @@
 package monocle.internal
 
-import monocle.Iso
+import monocle.{Iso, PIso}
 import scala.quoted.{quotes, Expr, Quotes, Type}
 import scala.deriving.Mirror
 
 object IsoFields {
-  transparent inline def apply[S <: Product](using mirror: Mirror.ProductOf[S]): Iso[S, Tuple] =
+  transparent inline def apply[S <: Product](using mirror: Mirror.ProductOf[S]): PIso[S, S, ? <: Tuple, ? <: Tuple] =
     ${ IsoFieldsImpl.apply[S]('mirror) }
 }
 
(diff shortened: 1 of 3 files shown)


julien-truffaut:
Hi @mbovel, sorry for the delay. I missed this PR.

Thanks for your doing it! would you mind adding the following?
1. Add a fallback match arm in `IsoFieldsImpl.apply` for better error messages:
```scala
  case other => report.errorAndAbort(s"Unexpected mirror type: ${other.show}")
```

2. adding a test for the extension syntax  in `IsoFieldsTest.scala`
```scala
  test("fields extension method works") {
    case class Foo(s: String, i: Int)
    val foo = Foo("hi", 5)
    val iso: Iso[Foo, (String, Int)] = foo.focus.fields  // or however the syntax works
    assertEquals(iso.get(foo), ("hi", 5))
  }
```

mbovel:
Thanks @julien-truffaut. I addressed your comments in 73c588b.

julien-truffaut:
Thanks @mbovel !

mbovel:
CI failed because of formatting; fixed, and rebased on top of main.

julien-truffaut (approved):

PR #1602, opened 2026-07-13
Remove whitebox cast from GenIso.fields

mbovel:
Fixes the `master` build, broken since #1574: the fallback arm of `GenIso._fields` passed the new wildcard-typed `IsoFields[S]` expression to `whitebox[A](e: Expr[Iso[S, A]])`, which no longer typechecks.

Rather than patching the cast, remove `whitebox` entirely and declare the honest supertype `PIso[S, S, ?, ?]`, as #1574 did for `IsoFields`. All match arms conform without a cast, and `transparent inline` still infers the precise type at call sites.

Verified with `sbt 'project rootJVM' '++ 3' test` (all tests pass, including `GenIso.fields` on 0-, 1- and 2-field case classes) and `scalafmtCheck`.

### DIFF
--- a/macro/src/main/scala-3/monocle/macros/GenIso.scala
+++ b/macro/src/main/scala-3/monocle/macros/GenIso.scala
@@ -1,6 +1,6 @@
 package monocle.macros
 
-import monocle.Iso
+import monocle.{Iso, PIso}
 import scala.language.`3.0`
 import scala.deriving.*
 import scala.quoted.*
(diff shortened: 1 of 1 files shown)


mbovel:
Ah, I didn't see #1600. I will rebase.

Removing `whitebox` entirely seems like a better fix to me. What do you think?

julien-truffaut:
Yes it does!

julien-truffaut (approved):

